PR #219, opened 2026-05-30
fix(tui/repl): bg/sub-agent notifications as standalone user messages

Leihb:
Background-process and sub-agent completion notifications previously rode the Steer path: wrapped in <system-reminder> blocks and folded into tool_result messages or prepended to the next turn. This made them easy to miss.

Now they are enqueued as standalone pending turns:
- **TUI**: bgExitMsg/subAgentNoteMsg enqueue to m.queue; idle auto-starts
- **Plain REPL**: uses a dedicated bgPending buffer + idleBgWait polling
- **formatBgNote/formatSubAgentNote** no longer wrap in <system-reminder>

This ensures notifications are treated as user messages, echoed into scrollback, and responded to explicitly by the model.

Fixes: user feedback that system-reminder notifications were being ignored.

### DIFF
--- a/cmd/octo/bgnotify.go
+++ b/cmd/octo/bgnotify.go
@@ -7,14 +7,12 @@ import (
 	"github.com/Leihb/octo-agent/internal/tools"
 )
 
-// formatBgNote renders a background-process completion as a <system-reminder>
-// block. It rides the existing steer path (Agent.Steer → folded into the next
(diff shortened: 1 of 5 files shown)


Leihb:
Closing: the approach was wrong. Will re-implement with steer path + plain text (no system-reminder wrap).

PR #538, opened 2026-06-10
feat(skills): add cron-task-creator default skill

Leihb:
## Summary
- Adds `internal/skills/defaults/cron-task-creator/SKILL.md`, a new default skill that guides the agent through configuring octo's built-in scheduled cron tasks (`~/.octo/tasks/*.json`, executed by the `octo serve` scheduler). No code changes — `go:embed defaults` picks the directory up automatically.

## What the skill encodes (verified against current code)
- **6-field cron syntax**: the scheduler is built with `cron.WithSeconds()` (`internal/scheduler/scheduler.go`), so standard 5-field crontab lines are invalid; the skill teaches the seconds-first format plus `@every`/`@daily` descriptors.
- **API-first creation** against a running `octo serve` (`POST /api/tasks` etc., `internal/server/tasks_handlers.go`), with a direct `~/.octo/tasks/<id>.json` file-write fallback when the server is down.
- **Operational caveats**: tasks only fire while `octo serve` runs; the scheduler arms lazily (first task-endpoint hit); edits/disables/deletes of existing tasks don't reschedule the live cron entry until a serve restart — only new `POST`ed tasks take effect immediately.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/skills/` (materialize + discovery mechanism)
- [x] `gofmt -l .` clean

### DIFF
--- a/internal/skills/defaults/cron-task-creator/SKILL.md
+++ b/internal/skills/defaults/cron-task-creator/SKILL.md
@@ -99,16 +99,11 @@ curl -s -X PATCH  http://127.0.0.1:8080/api/cron-tasks/{name} \
 
 - **Tasks only fire while `octo serve` is running.** No daemon, no serve → no
   runs. Missed schedules are not replayed on restart.
-- **The scheduler arms lazily.** After (re)starting `octo serve`, schedules are
-  not active until something touches a task endpoint — make one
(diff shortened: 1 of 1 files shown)


Leihb:
Updated the Caveats section to describe post-#541 behavior (live rescheduling, scheduler armed at serve start). Merge #541 first so the skill text matches the binary.

PR #953, opened 2026-06-30
feat(loop): unified anti-leak max-runtime cap (TUI/web/IM)

Leihb:
Follow-up to #951. Adds a single, unified safety cap so an in-session loop can't tick forever — the leak risk is mainly server-side (web/IM), where a forgotten loop keeps spending tokens with no one watching.

## One policy, three surfaces
- `tools.MaxLoopLifetime` (6h) + `tools.LoopExpired(start)` — one constant, one helper, referenced by every surface so the bound is identical on TUI, web, and IM.
- Each loop stamps a **start time on its first arm and keeps it across ticks**. Crucially, the spent-timer cleanup no longer resets it, so a **dynamic** loop's lifetime accumulates even though the model re-arms each turn (not just interval mode). Once past the bound, the loop **stops instead of re-arming**.
- Explicit stop (Ctrl+C / `schedule_wakeup(cancel=true)` / `/stop` / context reset) resets the clock.

## Wiring
- **Server**: `wakeupStart` map + `stopWakeupLocked`; `armWakeupFn` enforces the bound and `slog`s the stop. The fired timer callback now clears only the spent timer (keeps the clock).
- **TUI**: `loopStart` field + `wakeupFired` (keep-clock-on-tick); `armWakeup` enforces the bound and prints a notice.
- **Skill**: documents the cap (and reminds the model not to rely on it — stop loops when done).

## Tests
Symmetric lifetime-expiry tests on both surfaces, plus a "dynamic tick keeps the clock" assertion. Full suite green under `-race`; `gofmt`/`vet` clean.

6h is a named constant — easy to tune.

### DIFF
--- a/cmd/octo/loop_test.go
+++ b/cmd/octo/loop_test.go
@@ -3,6 +3,8 @@ package main
 import (
 	"testing"
 	"time"
+
+	"github.com/Leihb/octo-agent/internal/tools"
 )
 
 // armWakeupMsg (posted by the schedule_wakeup tool) arms the loop timer.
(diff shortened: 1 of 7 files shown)


Leihb:
Added lifecycle-hardening commit from an isolated code review:
- **Shutdown** stops all loop timers before teardown (no callback races adapter/session teardown).
- **Session delete** cancels the wakeup (an interval loop on a deleted session would otherwise re-arm forever).
- **Clock-leak fix**: `wakeupStart` is reclaimed after an abandoned dynamic loop's final turn, so the anti-leak clock map can't grow unbounded.
- New tests: actual fire + interval re-arm, dynamic-fires-once, idle clock reclamation, stopAll.

Rejected two review notes with reasoning: the `imWaker` capture matches the existing bg-completion idle-turn pattern (not a new defect), and cancelling the wakeup on turn-context cancel would break the coexist invariant.

PR #1099, opened 2026-07-03
fix(workflow): don't kill legitimate workflow_status checks; fix MCP stderr leak in one-shot

Leihb:
## Problem

Starting a background workflow in a headless one-shot produced a guaranteed dead end (screenshot in the session that found this):

1. The `workflow` start result said "call workflow_status to check progress" — actively inviting polling, and never mentioning that a completion notification is already pushed through the inbox (`FormatWorkflowNote`, wired in all transports).
2. `workflow_status` was missing from the stuck detector's `observationToolNames` exemption, so 5 identical status checks of the same `run_id` killed the turn with "repeated tool calls without progress" — while the workflow was, in fact, progressing (the detector fingerprints inputs only).
3. Comments in `repl.go`/`tuirepl.go` referenced `idleInboxWait`, which no longer exists.

Separately, the headless one-shot path passed `childStderr = nil` to `mcp.ConnectAll`, so stdio MCP subprocess diagnostics (e.g. `[CodeGraph MCP] File watcher active…`) leaked straight into octo's terminal output.

## Fix

- Exempt `workflow_status` in `observationToolNames` (same class as `terminal_output`).
- Add an anti-poll guard on `workflow_status` itself, mirroring `terminal_output`'s empty-snapshot guard: ≥3 consecutive still-running reads of one run append a hard `[STOP: …]` reminder; reading a finished run resets the counter (`WorkflowManager.RecordStatusRead`).
- Rewrite the `workflow` start result + both tool descriptions + the running-status detail text to promise the automatic completion notification and forbid polling, AsyncModeNotice-style. Also fixed the two remaining "poll workflow_status" invitations in `workflow_manager.go`.
- Remove the stale `idleInboxWait` comments; the `repl.go` one now documents the real one-shot behavior (no idle wait; exit kills running workflows).
- Route headless one-shot MCP child stderr to `~/.octo/logs/mcp.log` (the TUI's existing sink) instead of `os.Stderr`.

Not addressed here (follow-up under discussion): in one-shot mode a background workflow still cannot deliver its result if the model ends the turn — the process exits and kills the run. Direction being evaluated: run `workflow` foreground-blocking in one-shot.

## Tests

- `TestAgent_Run_RepeatedWorkflowStatus_DoesNotStop` — stuck detector no longer fires on repeated status checks.
- `TestWorkflowStatus_AntiPollingGuard` — STOP reminder appears at the threshold, absent before it, and absent on finished-run reads.
- `go test -race ./...` green; `gofmt -l` clean; `go vet` clean.

### DIFF
--- a/cmd/octo/chat.go
+++ b/cmd/octo/chat.go
@@ -776,35 +776,25 @@ func runChat(args []string, stdin io.Reader, stdout, stderr io.Writer) int {
 			fmt.Fprintf(stderr, "octo: mcp config: %v\n", err)
 		} else if len(mcpCfg.Servers) > 0 {
 			info := mcp.Implementation{Name: "octo", Version: version.Version}
+			// A stdio server's subprocess writes diagnostics to its stderr at
+			// arbitrary times, from an exec copy goroutine. Under the bubbletea
+			// TUI a direct terminal write corrupts the frame; in the headless
+			// one-shot it interleaves with the turn's own output (e.g. a
+			// file-watcher banner mid-turn). Route it to a log file in both
+			// modes (recoverable, never the screen). Connection warnings still
+			// reach the terminal via the warn writer. The file is closed when
+			// runChat returns.
+			var childStderr io.Writer
+			if f := openMCPLogFile(); f != nil {
+				childStderr = f
+				defer f.Close()
+			} else {
+				childStderr = io.Discard
+			}
 			if useTUI {
-				// A stdio server's subprocess writes diagnostics to its stderr at
-				// arbitrary times during the session, from an exec copy goroutine.
-				// Under the bubbletea TUI a direct terminal write corrupts the
-				// frame, so route it to a log file (recoverable, never the
-				// screen). The file is closed when runChat returns, i.e. after
-				// runTUI exits.
-				var childStderr io.Writer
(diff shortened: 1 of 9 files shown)


Leihb:
Isolated code review ran on this diff. Applied in adf5039:

- **Anti-poll streak now resets on observed progress** (`RecordStatusRead` compares `LastActivity` across reads) — spaced-out, user-prompted checks of a live run no longer escalate; only no-progress tight loops do, matching `terminal_output`'s empty-snapshot semantics.
- **List form guarded too** — the bare `workflow_status` list previously bypassed both the stuck detector (name-based exemption) and the per-run guard.
- Hoisted the duplicated `childStderr` block in `chat.go`; asserted the "DO NOT poll" start text; added a `RecordStatusRead` unit test (streak/reset/isolation) and a list-form guard test on an isolated manager.

**Deferred by design**: the review flagged that the "system will automatically notify you" promise is unreliable in headless one-shot (ending the turn kills the run before the notification lands). Correct, but the agreed follow-up — running `workflow` foreground-blocking in one-shot mode — removes background semantics from that mode entirely, so a transport-conditional wording flag built now would be ripped out immediately. Tracked for the follow-up PR.

PR #1102, opened 2026-07-03
feat(workflow): run workflows foreground-blocking in headless one-shot

Leihb:
## Problem

Follow-up to #1099. In headless one-shot mode a background workflow was a designed dead end: the model's only two options were polling `workflow_status` (discouraged by the anti-poll guard, and previously killed by the stuck detector) or ending the turn — which exits the process and kills the run via `KillDefaultWorkflows()`, silently discarding the work. The completion notification promised by the tool texts could never arrive.

## Fix

In one-shot mode the `workflow` tool now **blocks until the script completes and returns the final result inline** — no run id, no polling, no notification to wait for. Interactive transports (TUI, web, IM) keep the background contract unchanged.

- `tools.SetWorkflowForeground(true)` is set by the one-shot path in `runChat` **before** `DefaultToolsFor` builds the tool list, so the tool description advertises the right contract (`FOREGROUND` vs `BACKGROUND` mode line, pinned by test).
- New `WorkflowManager.Wait(ctx, id)` blocks on the run's `finished` channel. Ctx cancellation (SIGINT) cancels the run, waits for in-flight sub-agents to unwind, and returns the ctx error.
- Foreground runs still go through the manager — progress events, `workflow_status` visibility, and the 4-run concurrency cap all apply — but skip the completion hook (the result already returned through the call; a notification would reach the model twice).
- `runOnce` streams progress lines to stderr while the call blocks (`  · <line>`, suppressed in quiet mode), so a long run isn't minutes of silence, and drops its now-dead completion-notification wiring.
- Result rendering reuses `formatRunDetail`: done runs return output + journal id + log; script errors return the fix-and-rerun guidance as a normal tool result (not a Go error).
- Docs: workflows guide (en/zh) now documents the transport split.

## Verification

- New tests: foreground returns result inline + fires no completion hook; script error surfaces as `[error]` detail; SIGINT interrupt unwinds the run and errors; the description mode line flips with the flag.
- `go test -race ./...` green; gofmt/vet clean.
- End-to-end on the built binary: `./octo "…run this workflow…"` — the call blocked 250ms, progress lines streamed live to stderr, and the final reply carried the script result inline. No polling, no stuck detector, no lost run.


Leihb:
Isolated code review ran on this diff. One Important finding, confirmed and fixed:

- **Compute-only runaway scripts were uninterruptible** — wazero ran without `WithCloseOnContextDone`, so cancellation was only observable inside host calls (`agent()` waits). A pure-Ruby busy loop ignored `workflow_kill` in every transport, and in foreground one-shot mode it hung the whole process with SIGINT trapped (Wait's cancel branch blocks on the unwind). Fixed by enabling `WithCloseOnContextDone(true)` + widening `Result.Canceled` to cover the module-teardown avenue. Regression test added (`TestRun_CancellationComputeOnly` — deadlocks without the fix), plus an end-to-end check: SIGINT during a deliberate `loop { i += 1 }` run now unwinds with "run interrupted: context canceled" and exits cleanly.

Also applied from the review: mode-neutral concurrency-cap error text, deduplicated `mgr.Start`, hardened the no-notification assertion against goroutine ordering, `Wait`-on-unknown-run test.

Declined: routing progress lines through the view (`plainView.Notice` is the same bare `Fprintln(stderr)` as the direct write — no coordination gained).

PR #1164, opened 2026-07-05
fix(channel): fix inconsistent group-chat @-mention gating

### LINKED ISSUE #1122: im: group-chat gating inconsistent — DingTalk replies to any message containing "@", WeCom has no mention gate at all
## Problems

| Platform | Group gate |
|---|---|
| Telegram | real mention/reply required (`telegram.go:712`) |
| Discord | real mention required (`discord.go:706-715`) |
| Feishu | real mention required (`feishu.go:488-497`) |
| **DingTalk** | `if !mentioned && !strings.Contains(content, "@")` (`dingtalk.go:576-579`) — fires on ANY message containing an "@" character, including "@someone-else 帮我看下" — **the bot barges into unrelated group conversations** |
| **WeCom** | no gate: every message from an allowlisted user is processed regardless of chat type (`wecom.go:785-809`) |
| Weixin | DM-only (`weixin.go:711`) |

Also: a Telegram message from a non-allowlisted user is dropped with total silence (`telegram.go:576-578`) — defensible as anti-spam, but worth one first-contact hint (or at least a log) so legitimate users know they need allowlisting.

## Proposed fix

DingTalk: use the webhook's mention metadata (it carries the mentioned staff IDs) instead of a substring match. WeCom: require an @-mention in group contexts like the other adapters. Document the allowlist-silence behavior.

Leihb:
## Summary

Fixes #1122: DingTalk and WeCom's group-chat gates let the bot barge into
conversations it wasn't actually part of.

- **dingtalk.go**: the group gate was `!mentioned && !strings.Contains(content, "@")`
  — a message wasn't dropped if it merely contained an "@" character
  anywhere, e.g. "@someone-else 帮我看下". `atUsers` already carries the
  precisely-mentioned staff IDs from the webhook (the `mentioned` loop
  already existed), so the substring fallback was pure liability with no
  correctness upside. Removed it; the gate is now `atUsers`-only, matching
  how Telegram/Discord/Feishu already do it via their own mention metadata.

- **wecom.go**: there was no group gate at all — every message from an
  allowlisted user was processed regardless of chat type. Verified against
  WeCom's own aibot docs before implementing: their sample shows group
  messages arrive as literal `"@RobotA hello robot"` text, with no
  structured "was the bot mentioned" field to check instead (unlike
  DingTalk's `atUsers` or Telegram/Discord/Feishu's entity metadata). Since
  matching that text requires knowing the bot's own display name, and
  nothing in config carried it, added `bot_nickname` as a new **optional**
  config key: when set, a group message must contain `"@<nickname>"` to
  pass; when unset, group gating stays off (matches pre-fix behavior, so
  existing deployments don't go silent in every group they're already in).

Not changed: the issue's aside about a non-allowlisted Telegram user's
silent drop being worth a first-contact hint — that's a separate UX
enhancement, not part of the gating-inconsistency bug.

## Verification
- [x] `go build ./...`, `go vet ./...`
- [x] `go test -race ./...` (full suite)
- [x] New tests: DingTalk mention-required vs. real-mention-passes; WeCom
      no-nickname-configured (ungated, unchanged), nickname-configured group
      gating, and single-chat never gated regardless of nickname config.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/internal/channel/adapters/dingtalk/dingtalk.go
+++ b/internal/channel/adapters/dingtalk/dingtalk.go
@@ -590,7 +590,12 @@ func (a *Adapter) handleInbound(raw json.RawMessage, onMessage func(channel.Inbo
 	}
 	a.routeMu.Unlock()
 
-	// Group chat: only @-mention.
+	// Group chat: only @-mention. #1122: this used to also pass on any
+	// message containing a bare "@" character (e.g. "@someone-else 帮我看下"),
+	// which let the bot barge into unrelated group conversations. AtUsers
+	// already carries the precisely-mentioned staff IDs from the webhook, so
+	// there is no need for — and no correctness reason to keep — the
+	// substring fallback.
 	if ev.ConversationType == "2" {
 		mentioned := false
 		for _, u := range ev.AtUsers {
@@ -599,8 +604,7 @@ func (a *Adapter) handleInbound(raw json.RawMessage, onMessage func(channel.Inbo
 				break
 			}
 		}
-		content := ev.Text.Content
-		if !mentioned && !strings.Contains(content, "@") {
+		if !mentioned {
 			return
 		}
 	}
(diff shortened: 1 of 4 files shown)


Leihb:
The Svelte Check CI step has failed, though the Go checks are still running. The PR description mentions `go test -race ./...` passed locally — the failure may be in the Svelte/lint workflow unrelated to the Go changes. Could you take a look at the [Svelte Check logs](https://github.com/open-octo/octo-agent/actions/runs/28728854512/job/85190982170) to see what needs fixing?

PR #1185, opened 2026-07-05
fix(tools): anchor glob's walk to the pattern's literal path prefix

Leihb:
## Summary
- `glob`'s file enumeration always walked the entire search root via `rg --files` before filtering matches in Go, so every call cost the same full-tree scan regardless of how specific the pattern was (e.g. `cmd/octo/config.go` cost the same as `**/*`). `grep` already hands its `include` glob to ripgrep's native `--glob` pruning; `glob` never did the equivalent.
- Added `literalPathPrefix(pattern)` to extract the pattern's leading non-wildcard path segments (e.g. `internal/tools` from `internal/tools/**/*.go`) and anchor the walk there instead of the whole root. A fully literal pattern with no wildcards resolves straight to a single file (O(1) via the existing non-directory-root path in `listProjectFiles`).
- When the literal prefix doesn't exist on disk at all, skip the `rg` subprocess entirely and return "no matches" directly.
- Sorting/truncation (stat + mtime sort before capping to 200) is unchanged — that's required for correct recency ranking, not part of the bug.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/tools/...` (all existing `TestGlob_*` pass unchanged)
- [x] New: `TestLiteralPathPrefix` — table test of the prefix-extraction helper
- [x] New: `TestGlob_PrunesToLiteralPrefix` — pattern anchored to one subdir + an unreadable sibling dir asserts no `[warning:` surfaces, proving the walk never touched the sibling (would have before this fix)
- [x] New: `TestGlob_LiteralPrefixMissingSkipsWalk` — nonexistent literal prefix returns "no matches" without needing `rg` on PATH at all
- [x] New: `TestGlob_FullyLiteralPatternMatchesSingleFile` — fully literal pattern matches exactly one file, not siblings

### DIFF
--- a/internal/tools/glob.go
+++ b/internal/tools/glob.go
@@ -97,7 +97,27 @@ func (GlobTool) Execute(ctx context.Context, _ string, input map[string]any) (ag
 		candidate := filepath.Join(absRoot, prefix)
 		switch _, statErr := os.Stat(candidate); {
 		case statErr == nil:
-			scanRoot = candidate
+			// os.Stat resolves case-insensitively on the filesystems most
+			// developers actually run this on (macOS APFS, Windows NTFS,
+			// both case-insensitive by default) — so a pattern whose
+			// casing doesn't match what's on disk (e.g. "Internal/*.go"
+			// against a real "internal/") would otherwise get pruned to
+			// that wrong-case candidate anyway, and rg would then report
+			// matches under the wrong-case path. glob's matching is
+			// documented as path.Match semantics, i.e. case-sensitive —
+			// on a case-sensitive filesystem (Linux) that mismatch would
+			// already fail the stat and never reach here, so only trust
+			// the prune once we've confirmed the same case-sensitive
+			// result by checking each segment's on-disk name byte-for-byte.
+			if onDiskCaseMatches(absRoot, prefix) {
+				scanRoot = candidate
+			}
+			// else: leave scanRoot as absRoot — the prune target only
+			// exists via a case-insensitive lookup, so pruning to it would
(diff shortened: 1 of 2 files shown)


Leihb:
Ran an isolated code review on this. Found two real regressions in the pruning logic, both confirmed by repro and now fixed in the latest commit:

1. **Critical — path traversal via `..`**: `literalPathPrefix` treated `..` as an ordinary literal segment, so `pattern="../sibling/*.go"` pruned the walk to `filepath.Join(root, "../sibling")` — outside the intended search root. Repro confirmed files from the parent/sibling directory leaked into results. Before this PR's pruning existed, that pattern could never match anything (no real path has a literal `..` component), so the fix just stops prefix collection at `..`, restoring the "no matches" behavior.
2. **Important — case-insensitive-filesystem mismatch**: `os.Stat` resolves case-insensitively on macOS/Windows defaults, so a case-mismatched prefix (e.g. `Root/*.go` against a real `root/`) passed the stat check and pruned to the wrong-case path — silently making glob's documented case-sensitive matching platform-dependent. Repro confirmed a mismatched-case pattern matching when it shouldn't. Fixed by verifying each prefix segment against the real directory entry name (`onDiskCaseMatches`) before trusting the prune.

Both have regression tests now (`TestGlob_DotDotPatternNeverEscapesRoot`, `TestGlob_CaseMismatchedPrefixDoesNotMatch`), plus extra `literalPathPrefix` table cases for `..`, leading `/`, and `./`. Full suite (439 tests) still green.

PR #1254, opened 2026-07-08
fix(tui): say "Goal starts" on the first turn, not "Goal continues"

Leihb:
## Summary
- `goalContinuationKick()` shared one notice ("● Goal continues — /goal pause to stop") between the very first kickoff turn (`startGoalNow`, fired right after `/goal create/replace/resume`) and every subsequent auto-continuation turn (`handleTurnFinished`). The first turn hasn't continued anything yet, so it read wrong.
- `goalContinuationKick` now takes an `initial bool`: `true` (from `startGoalNow`) prints "Goal starts", `false` (from `handleTurnFinished`) keeps "Goal continues".

## Test plan
- [x] `go build ./...`
- [x] `go test ./cmd/octo/...` (added `TestGoalContinuationKick_InitialAnnouncesStart` / `...FollowupAnnouncesContinues`)

### DIFF
--- a/cmd/octo/tuirepl.go
+++ b/cmd/octo/tuirepl.go
@@ -2039,7 +2039,7 @@ func (m *tuiModel) handleTurnFinished(err error) (tea.Model, tea.Cmd) {
 	// the policy (status, zero-progress suppression); user input always won
 	// above by reaching the queue/inbox branches first.
 	if err == nil {
-		if prompt, ok := m.goalContinuationKick(); ok {
+		if prompt, ok := m.goalContinuationKick(false); ok {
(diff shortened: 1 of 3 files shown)


Leihb:
Self-review catch: the original diff hardcoded `initial: true` at every `startGoalNow()` call site, including `/goal resume`. Resume picks the *same* goal back up after `applyGoalStatus` already printed "Goal active — ...", so it isn't a fresh start — it should still say "continues". Pushed a follow-up commit (cbbbb05) threading a `fresh bool` through `startGoalNow`'s four call sites: `true` for create/replace (a brand-new goal), `false` for resume (matching `handleTurnFinished`'s auto-continuation).

Re-verified: `go build ./...` and `go test ./cmd/octo/...` pass.

PR #1383, opened 2026-07-12
feat(trash): correctness pass + session-title labels

Leihb:
First of a staged upgrade to the file recycle bin (回收站). This is **PR 1 of 4**; it lands the **safety fixes** and the **recognizability fix**. Overwrite protection, CLI/TUI access + inline undo, and retention/eviction follow in later PRs. Full design: `dev-docs/trash-recall-design.md`.

## What this fixes

- **Unique entry IDs** — `ID` is now `<project_hash>.<trash_name>`, a single URL-safe segment that pins the exact file. Previously `Restore`/`Delete` scanned all projects and first-matched by basename, so two projects deleting `config.json` could restore the wrong file to the wrong path. A random token in the staged name also prevents same-second same-basename collisions within one project.
- **Restore never silently overwrites** — the common case (deleted a file, recreated it, now want the old one back) used to `os.Rename` straight over the current file. `Restore(id, policy)` now defaults to `ConflictAbort` → `ErrRestoreConflict`; the web handler replies `409`. `ConflictBackupExisting` (stage the current file first) and `ConflictRename` (`<name>.restored-<ts>`) cover resolution. Cross-filesystem safe (rename → copy+remove).
- **Atomic Move** — renames into the trash when same-filesystem (instant, space-free even for big trees), writes meta after, rolls back on meta failure; copy+remove fallback across filesystems.
- **POSIX `rm` wrapper: absolute-path bug** — it copied from `$PWD/$arg` unconditionally, so `rm /abs/path` staged nothing (the copy source didn't exist) **yet the real `rm` still deleted it** — absolute-path deletes were silently unprotected. Now uses the already-resolved absolute path as the copy source. Also hard-links into the trash when possible (a `node_modules`-sized tree stages instantly) and adds entropy to the staged name.
- **Session-title labels** — `Entry.Label` surfaces a session transcript's title in listings, parsed from the trashed `.jsonl` (first `meta` line, then a later `title` record; a `type` guard ignores web-search page titles embedded in message content). A trash full of `20260712-…​.jsonl` files is now legible; the web view leads with the label.

## Verification

- `go build/vet ./...` clean, `gofmt -l` clean, `go test -race ./internal/{trash,tools,server}` green.
- New unit tests: cross-project unique IDs, all three restore conflict policies, no-conflict restore (recreates parent dir), directory Move, Delete, session-title parsing (meta + title-record override + non-session empty), and the **absolute-path `rm` regression** run through the real POSIX wrapper via `sh`.
- Also smoke-tested `octo serve` + `GET /api/trash` against a populated local trash to confirm session labels resolve end-to-end and non-session/titleless entries fall back to the basename.


Leihb:

PR #1386, opened 2026-07-12
feat(trash): back up files before write_file/edit_file overwrites

Leihb:
**PR 2 of 4** in the recycle-bin (回收站) upgrade. Closes the biggest safety hole: overwriting an existing file — the agent's most common destructive action — bypassed the trash entirely (the package doc claimed overwrites were covered; only `rm` and programmatic deletes were). Depends conceptually on #1383 but touches disjoint files. Design: `dev-docs/trash-recall-design.md` (lands with #1383).

## What this adds

`write_file` and `edit_file` now stage the current bytes into the trash **before** clobbering, via a new best-effort `backupBeforeOverwrite`. A git-aware gate keeps the trash useful rather than noisy:

| File state | Action |
|---|---|
| Untracked | **back up** — nothing else can recover it |
| Tracked but dirty | **back up** — working tree differs from anything git holds |
| Tracked and clean | **skip** — `git checkout` recovers the exact bytes |
| Not a git repo / git missing | **back up** |

- Best-effort, mirroring the safe-rm wrapper: a backup failure never blocks the write.
- The git check runs **only when overwriting an existing file**, so creating a new file pays nothing.
- Default on; opt out with `trash.overwrite_backup: false` (new `TrashConfig`).

## Test isolation

Extends the tools package `TestMain` (which already redirects the workflow dirs away from real `~/.octo`) to also point `HOME` at a throwaway dir — otherwise the new overwrite path would stage test files into the developer's real `~/.octo/trash` during `make test`. The temp HOME is created **under the real home dir, not TMPDIR**, because the sandbox integration tests write a probe to `$HOME` expecting the OS sandbox to deny it, and macOS's default policy allows TMPDIR (a temp HOME there would let the probe through and break them).

## Verification

- `go build/vet ./...` clean, `gofmt -l` clean.
- `go test -race ./internal/{config,tools}` green (one unrelated pre-existing flake: `TestBrowserTool_CookiesIncludesHttpOnly`, a live-Chrome CDP test — passes on its own).
- New tests cover: untracked → staged, dirty-tracked → staged, clean-tracked → skipped, create → no-op, toggle-off → no-op.
- Confirmed the change no longer pollutes the real `~/.octo/trash` (before/after project-dir count unchanged across a full package run).

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -138,6 +138,25 @@ type Config struct {
 	// Language persists the user's UI language preference ("en" | "zh").
 	// Empty means the default (English).
 	Language string `yaml:"language,omitempty"`
+	// Trash configures the file recycle bin (agent-issued deletes and
+	// overwrites are staged there for recovery).
+	Trash TrashConfig `yaml:"trash,omitempty"`
+}
+
+// TrashConfig configures the file recycle bin.
+type TrashConfig struct {
+	// OverwriteBackup stages a copy of a file into the trash before write_file
+	// or edit_file overwrites it, so an overwrite that destroys uncommitted
+	// work is recoverable. Files that are tracked by git and clean are skipped
+	// regardless — git already holds a recoverable copy. nil means the default
+	// (enabled).
+	OverwriteBackup *bool `yaml:"overwrite_backup,omitempty"`
+}
+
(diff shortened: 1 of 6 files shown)


Leihb:

PR #1389, opened 2026-07-12
feat(trash): reach the recycle bin from the terminal (octo trash + /trash)

Leihb:
**PR 3 of 4** in the recycle-bin (回收站) upgrade. The trash was only reachable from the Web UI — a recovery feature the *primary* interface (CLI/TUI) couldn't use. This adds terminal access.

> **Stacking:** depends on #1383 (trash API: conflict-policy restore, `Delete`, `Entry.Label`, unique IDs) and #1386 (overwrite backup). Since those are parallel branches off `main`, this PR's diff currently includes their commits too — **review commit `a7edf982`**. Once #1383 and #1386 merge, I'll rebase and the diff collapses to just this change. No auto-merge.

## Commands

```
octo trash [list]                    list recoverable files, newest first
octo trash restore <id|substring>    restore to the original path
      --overwrite                      trash the current file first, then restore
      --as-copy                        restore as <name>.restored-<ts>
octo trash rm <id|substring>         permanently delete one entry
octo trash empty --all|--old|--orphans
```

- `restore`/`rm` resolve an exact ID **or a unique substring** of the id, original path, or label — no need to copy a full timestamped id. An ambiguous match lists the candidates instead of guessing.
- Conflicts surface the package's safe policy: restoring onto an occupied path **aborts with a hint** toward `--overwrite` / `--as-copy` rather than clobbering.
- Listings lead with the human label (session title) when present, falling back to the basename.

## TUI

`/trash` (with completion + help entries) reuses the **exact** CLI logic — `list`, `restore` (incl. `--overwrite`/`--as-copy`), `rm`, and `empty` all work in-REPL and behave identically to the standalone command.

## Scope note

Inline "↩ undo" in the transcript is **not** in this PR — it touches the tool result UI plus both the TUI and web renderers, so it's folded into the next PR alongside the web provenance display and retention/eviction. This PR is the self-contained "terminal reach" slice.

## Verification

- `go build/vet ./...` clean, `gofmt -l` clean, `go test -race ./cmd/octo` green.
- New tests: list (empty + populated), restore by substring, restore-conflict aborts then resolves with `--overwrite`, rm by substring, ambiguous-substring lists candidates, `empty` requires a mode, unknown subcommand exits 2.
- Confirmed a full `cmd/octo` test run leaves the real `~/.octo/trash` unchanged (tests isolate HOME).


Leihb:

PR #1390, opened 2026-07-12
feat(trash): bound the recycle bin with retention + a size cap

Leihb:
**PR 4** in the recycle-bin (回收站) upgrade. The overwrite protection (#1386) stages a copy on every dirty-file edit, so the trash now grows on its own — it needs an upper bound.

> **Stacking:** builds on #1383 → #1386 → #1389. Until those merge this PR's diff includes their commits — **review commit `ef43d73f`**. I'll rebase to collapse it once they land. No auto-merge.

## What this adds

`trash.Enforce(maxBytes, retention)`:
1. Ages out entries older than the retention window.
2. If the total still exceeds the size cap, evicts **oldest-first** until back under.

Orphans (their original project gone) are **skipped by the age-out** — often exactly what a user wants back after a checkout moved a repo — but stay eligible for hard-cap eviction, since the cap is a firm limit.

Runs in a **background goroutine at startup** (`runChat` + `runServe`), best-effort: the recycle bin working never depends on housekeeping having finished. Both bounds are configurable, defaulting to **14 days / 10 GiB**:

```yaml
trash:
  retention_days: 14      # 0 = default; negative disables age-out
  max_size_mb: 10240      # 0 = default (10 GiB); negative disables the cap
```

## Test hygiene

A test-only `init()` disables the startup housekeeping in the `cmd/octo` test binary, so `go test` can never evict the developer's real `~/.octo/trash` (several tests drive `runChat`/`runServe`).

## Verification

- `go build/vet ./...` clean, `gofmt -l` clean, `go test -race ./internal/{config,trash} ./cmd/octo` green.
- New tests: age-out removes old / keeps recent, **orphan survives age-out**, size cap evicts oldest-first to just under the limit, both-bounds-disabled is a no-op; config accessors for defaults / explicit / negative-disables.
- Confirmed a full `cmd/octo` run leaves the real `~/.octo/trash` unchanged.


Leihb:

PR #1401, opened 2026-07-13
Teach code-review skill to dispatch sub-agents in parallel

Leihb:
The code-review skill instructions told the caller to dispatch an isolated sub-agent but did not explain how to run multiple independent reviews concurrently. In practice, a review of several PRs would block on the first sub-agent and only start the second after it returned.

Add a short paragraph to the "Dispatch the sub-agent" step explaining:

- For multiple independent reviews, set `run_in_background: true` on each `sub_agent` call.
- Collect the results from completion notifications or via `sub_agent_status`.
- A single review can still use a synchronous `sub_agent` call.


Leihb:
Superseded by #1402, which includes this skill fix plus the broader sub_agent tool description updates.

PR #1414, opened 2026-07-13
fix(tui): reject /model switch to unconfigured model with a clear error

Leihb:
## Problem

Previously, `/model <name>` against a model not present in the config silently fell back to the current provider. When switching from e.g. LongCat-2.0 to `deepseek-v4-pro` (not configured), `resolveProviderModel` returned `flagProvider` (the active one) because `EntryByModel` missed. `ensureSender` then saw no provider change and kept the old sender bound to the wrong endpoint — requests to `api.longcat.chat/openai` with a deepseek model name → **HTTP 500**.

## Fix

`ensureSender` now pre-checks `EntryByModel` before falling through to provider resolution. If the target model is not in the config, it returns an error listing the configured models and leaves cfg untouched (session stays on the old model).

## Test

Added `TestEnsureSender_UnconfiguredModel` — verifies the error path leaves providerName, configEntry, and sender unchanged. All 6 ensureSender tests + full `cmd/octo` package pass with `-race`.

### DIFF
--- a/cmd/octo/ensure_sender_test.go
+++ b/cmd/octo/ensure_sender_test.go
@@ -213,3 +213,43 @@ func TestEnsureSender_ErrorLeavesConfigUnchanged(t *testing.T) {
 		t.Error("sender should be unchanged after failed rebuild")
 	}
 }
+
+// TestEnsureSender_UnconfiguredModel verifies that switching to a model not
+// present in the config returns a clear error (listing what is configured)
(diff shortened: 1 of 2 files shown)


Leihb:
Superseded by #1416 — the unconfigured-model guard (cherry-picked from this PR) is now merged as part of the larger  picker feature. Closing in favor of that.

PR #1428, opened 2026-07-13
feat: add Snip tier before LLM summarize in compaction pipeline

kackpack:
## Summary

The compaction pipeline has two tiers: reclaim (elide stale tool results, zero cost) and summarize (LLM call). This adds a Snip tier between them — when context still exceeds the compaction trigger by 50%+ after reclaim, Snip drops the oldest safe user turn at zero LLM cost.

## How it works

Snip runs inside `maybeCompact()`, after reclaim and before summarize. It looks for the oldest complete user turn that can be safely removed — one whose messages contain no error tool results and whose removal doesn't reach into the first turn (msg[0-1], which carries the task definition). If found, Snip drops that turn from history and returns, skipping the summarize call entirely. If no safe candidate exists, or if context is less than 50% past the trigger, Snip does nothing and the pipeline falls through to summarize.

## When it triggers

- **Context exceeds trigger by 50%+** — for a 100K-token trigger, Snip fires at ~150K. Below that, summarize handles it.
- **Reclaim alone wasn't enough** — reclaim runs first; Snip only attempts if context is still over.
- **Trigger itself is ≥ 1000 tokens** — a safety floor to avoid acting on artificially low test thresholds.

## Changes

- `internal/agent/compaction.go` — +76 lines: `snipBeforeSummarize()` function + call in `maybeCompact()`
- `internal/agent/event.go` — +17 lines: `EventSnip` constant + `SnippedTurns` field on `CompactStats`
- `internal/agent/snip_test.go` — +95 lines: 5 tests

## Design decisions

- **50% threshold**: deliberately conservative. Snip only fires when context is severely overshot; mild overshoot falls through to summarize. Prevents thrashing where Snip fires on every turn.
- **snipMinTriggerTokens = 1000**: floor to avoid firing on artificially low thresholds used in unit tests.
- **No archiving**: Snip trades recoverability for zero latency; the auto-compact archiver only runs for summarize folds.

## Guards

- **msg[0-1] preserved** — task definition and system constraints
- **IsError=true turns preserved** — error context is critical for recovery
- **Whole-turn boundaries only** — tool_use/tool_result pairs never severed

## Edge cases (known)

- Enormous msg[0]: Snip's guard keeps it; falls back to summarize.
- Large error tool results: existing `microCompact` (40KB per-result cap) truncates them before Snip sees the turn.
- Manual `/compact` also goes through the Snip tier before summarize.

---

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -26,6 +26,18 @@ const minFoldPercent = 15
 // kept tail, the summary side-call, and the next turn's output.
 const compactThresholdFraction = 0.75
 
+// snipOvershootThresholdFraction is how far past the trigger context must
+// sit for Snip to fire. A 50% margin ensures Snip only fires when summarise
+// would clearly be wasteful — the context is severely over the trigger and
+// dropping a few stale turns is far cheaper than an LLM call.
+const snipOvershootThresholdFraction = 0.50
+
+// Snip only fires when the compaction trigger itself is above this floor.
+// Unit tests often set tiny artificial thresholds (e.g. 100 tokens) to
+// exercise the summarise path — snipping at that scale isn't meaningful
+// and would break the test expectations.
+const snipMinTriggerTokens = 1000
+
 // defaultContextWindow is the conservative fallback window (in tokens) for
 // models not named in contextWindow. Under-estimating only makes us compact
 // slightly earlier — never overflow — so unknown models stay safe.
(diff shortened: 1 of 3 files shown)


Leihb:
Thanks for this — genuinely nice work. Clear write-up, guards spelled out, tests and event/UI plumbing all included, and picking up the compaction pipeline (one of the more fragile paths in the codebase) is not the easy corner to
  contribute to. I appreciate it.

  Before we get into implementation, I want to raise a design-level question, because I think that's the thing to settle first — I'd rather not have you sink more effort into the code until we're aligned on whether this tier should exist
  in the shape proposed.

  A few concerns with Snip as a distinct tier:

  - It's an unrecoverable, un-archived delete. reclaim only sheds tool-output bodies while keeping the conversational skeleton, and summarize is lossy-but-preserving — it folds old turns into a summary and archives the verbatim originals
  for the model to read back. Snip drops whole turns with no summary and no archive. Recoverability is something we lean on as a feature elsewhere, so a tier that permanently discards context is a bit against the grain — especially for
  coding sessions, where the oldest turns often carry the task definition and key decisions.
  - What it saves is already cheap. The whole premise is "avoid the summarize call," but summarize is infrequent (only on threshold crossings) and already runs on the lite model when one is configured. The marginal cost being avoided is
  small.
  - At the trigger point, the bulk is usually in the tail — which Snip doesn't touch. Snip only fires at 50%+ overshoot, i.e. when reclaim couldn't bring things down. In practice that overshoot is typically driven by a recent, oversized
  agentic turn wedged in the kept tail (the same case reclaim's comment calls out). Snip drops the oldest turns and leaves the tail alone, so in exactly the situation that triggers it, it may drop old context and still not get under the
  threshold.

  So the question I'd like to answer first: what's the actual motivation — saving token cost, or removing the latency of the synchronous summarize call?

  - If it's cost: I don't think it clears the bar — the savings are small and the price (unrecoverable context loss + added complexity on a fragile path) is high.
  - If it's latency (the summarize call blocking the current turn): that's a real problem worth solving, but I'd rather solve it in a way that fits the pipeline's direction — e.g. making summarize async, or only snipping when it provably
  shrinks the context meaningfully — rather than adding a tier that permanently deletes turns.

  For completeness, there are also a couple of implementation issues that would need rework regardless of direction (in short: on a plain non-tool conversation the reconstruction is a no-op yet still returns early and skips summarize, so
  context never shrinks; and on an agentic first turn it can sever a tool_use/tool_result pair and 400 the next request). Happy to detail those — but I'd hold off until we've settled the design question above.

  Would love to hear which motivation is driving this — if it's latency, I think there's a good version of this we could land.

kackpack:
Thanks for the thorough review — you're spot on. Unrecoverable deletion totally goes against the pipeline's philosophy. Since Collapser (#1449) already solves the cache issue without being destructive, I'll go ahead and close this one. Cheers!

PR #1449, opened 2026-07-14
feat: replace reclaim ReplaceAll with read-time Collapser projection

kackpack:
## Summary

`reclaimStaleToolResults` currently calls `History.ReplaceAll`, which replaces the entire messages array — invalidating the provider's prompt cache prefix. This PR replaces that with a **Collapser** that records collapsed block positions in a lightweight map, and `Snapshot()` projects them at read time without mutating the history.

## Changes

- `internal/agent/collapser.go`: +79 lines — Collapser struct with set/get/clear/remove/project
- `internal/agent/collapser_test.go`: +133 lines — 8 tests
- `internal/agent/history.go`: +15 lines — Collapser field, Snapshot projection, collapser clear on Reset/TruncateTo/ReplaceAll
- `internal/agent/reclaim.go`: −30/+22 lines — rewrite to write Collapser entries instead of ReplaceAll

## How it works

Before:
```
reclaim → mutate snapshot copy → History.ReplaceAll(modified copy)
          → messages array pointer changes → prompt cache invalidated
```

After:
```
reclaim → History.collapser.set(msgIdx, blockIdx, placeholder)
          → messages array stays unchanged → prompt cache intact

Snapshot():
  copy messages → project via collapser entries → return projected copy
```

## Side effects

- Collapser is cleared on Reset / TruncateTo / ReplaceAll / replaceLast — no stale entries leak.
- Collapser is NOT persisted across session save/load — the cost is one turn of full visibility after restore (reclaim re-populates on next trigger).

## User-visible impact

---

## Self-review

### 1. Architecture

### 2. Need

### 3. Code

- [ ] All tests pass

### Bonus

---

### DIFF
--- a/internal/agent/collapser.go
+++ b/internal/agent/collapser.go
@@ -1,0 +1,79 @@
+package agent
+
+import "strconv"
+
+// Collapser records which tool_result blocks have been collapsed by
+// reclaimStaleToolResults so Snapshot() can project the placeholder at read
+// time without mutating the underlying messages array. This keeps the
+// provider's prompt cache prefix valid because the stored message slice is
+// never rewritten — only the returned copy carries the projection.
+//
+// Key format: "msgIdx:blockIdx" — two ints joined by a colon.  This is
+// deliberately a string key rather than a struct to keep the zero-value useful
+// (a nil map is empty and needs no initialisation).
+//
+// Collapser is NOT persisted across session save/load. After a restart
+// reclaim will re-evaluate and re-populate entries, so the only cost is one
+// turn of full tool-result visibility after restore — accepted tradeoff.
+type Collapser struct {
+	entry map[string]string // msgIdx:blockIdx → placeholder text
+}
+
+// set records a collapsed placeholder for message at mi, block at bi.
+func (c *Collapser) set(mi, bi int, placeholder string) {
+	if c.entry == nil {
+		c.entry = make(map[string]string)
+	}
+	c.entry[c.key(mi, bi)] = placeholder
+}
+
+// get returns the placeholder and true if a collapse entry exists.
+func (c *Collapser) get(mi, bi int) (string, bool) {
+	if c.entry == nil {
+		return "", false
+	}
+	s, ok := c.entry[c.key(mi, bi)]
+	return s, ok
+}
+
+// remove clears collapse entries for message mi (used when that message is
+// overwritten by replaceLast).
+func (c *Collapser) remove(mi int) {
+	if c.entry == nil {
+		return
+	}
+	prefix := strconv.Itoa(mi) + ":"
+	for k := range c.entry {
+		// Fast path: prefix matches
+		if len(k) > len(prefix) && k[:len(prefix)] == prefix {
+			delete(c.entry, k)
+		}
+	}
+}
+
+// clear drops all entries.
+func (c *Collapser) clear() {
+	c.entry = nil
+}
+
+// len reports the number of entries (0 when never used).
+func (c *Collapser) len() int {
+	return len(c.entry)
+}
+
+// project replaces collapsed tool_result blocks in-place on a copy of the
+// message slice (caller must own the copy — Snapshot handles this).
+func (c *Collapser) project(msgs []Message) {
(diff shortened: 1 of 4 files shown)


Leihb (commented):
Thanks a lot for taking the time on this, and for the unusually thorough write-up — the before/after diagrams, the side-effects section, and 8 tests made the intent very easy to follow. I appreciate the care.

That said, I don't think we should land this, because the problem it targets doesn't actually exist, and the change introduces a real data race. Details below.

## The core premise doesn't hold: there's no cache to save

The PR's motivation is that `ReplaceAll` invalidates the provider's prompt-cache prefix and a read-time projection avoids it. Anthropic's prompt cache matches on the **actual request bytes** (see `internal/provider/anthropic/client.go` — `markMessagesCacheable`/`markMessageCacheable` place breakpoints on the static system+tools prefix plus the last two messages; a hit is purely the longest byte-identical prefix of the request).

The bytes we send come from `a.History.Snapshot()` (`agent.go:553/617/...`). In **both** implementations:

- old: reclaim rewrites the stale block to a placeholder, `ReplaceAll` stores it → `Snapshot()` returns the placeholder
- new: reclaim records it in the collapser → `Snapshot()` projects the placeholder at read time

…the request bytes are identical: an old block shrunk to a placeholder. So the cache prefix breaks at exactly the same point (the first collapsed block) in both cases, byte-for-byte. Collapsing a stale block necessarily changes that message's content and invalidates the suffix — that's the inherent cost of reclaiming context, independent of how we write it back.

`ReplaceAll`/`rewritten` only affect **on-disk session persistence** (`session.go:1319-1320`; `takeRewriteDirty()` chooses append vs. full-file rewrite). They have no effect on the provider request or its `cache_control` markers. The old comment "invalidates the prompt-cache prefix from the first rewrite" refers to the **content change**, not to the `ReplaceAll` call itself.

Net: +79 lines of Collapser, changes to 5 `History` methods, and a new positional-key invariant — for zero cache benefit.

## Data race (blocking)

`History` is documented as goroutine-safe: every access to `messages`/`rewritten` goes through `h.mu`. But reclaim now does:

```go
a.History.collapser.set(l.mi, l.bi, placeholder)   // reclaim.go — holds no lock
```

while `Snapshot()` reads `h.collapser.project()/len()` under `RLock`, and `Reset/TruncateTo/ReplaceAll` call `h.collapser.clear()` under `Lock`. `set()` mutates the `entry` map with no synchronization against those readers/clearers — a genuine data race on the map. The old path was safe because folding landed via `ReplaceAll`, which takes `h.mu.Lock()` itself.

The included tests call reclaim and `Snapshot()` sequentially on one goroutine, so `go test -race` won't surface it — but the server (background save goroutine, WS handlers, run loop) will hit it concurrently.

## Latent: `popLast` doesn't maintain the collapser

`replaceLast` was updated to call `collapser.remove(...)`, but `popLast` (`agent.go:1899`, used by `Turn` to undo the user-message append on a Sender failure) drops the last message without touching the collapser. If a collapsed index is popped and a later `Append` reuses that index, the stale placeholder projects onto the wrong block. It's unlikely to fire in practice (reclaim only touches `locs[:len-hotToolResults]`, so the hot tail is never collapsed) — but it illustrates the cost of absolute `(msgIdx, blockIdx)` keys: every history mutation now has to remember to keep the collapser in sync, and missing one fails silently.

## Suggestion

If there's a concrete, measurable win I'm missing, I'd genuinely like to see the scenario — happy to reconsider. As it stands the cache goal isn't real, so I'd lean toward closing this. If we do want to pursue it, at minimum: (1) move all collapser access (`set/remove/clear/project`) behind `h.mu` via `History` methods, (2) add a `-race` concurrency test, and (3) handle `popLast`.

Thanks again for the contribution and the quality of the writeup — hope the detail above is useful even if the outcome isn't the one we'd hoped.

kackpack:
Thanks for the detailed explanation — I was wrong about how Anthropic’s prompt cache works. The Collapser doesn’t solve a real problem and introduces a data race. Appreciate the patience on both PRs.

PR #1464, opened 2026-07-14
fix(config): allow empty model name on the first entry as floating default

Leihb:
## Problem

After running `octo config` (or the first-run config wizard) and accepting a vendor's built-in default model, the saved `~/.octo/config.yml` has `models[0].model` set to an empty string. `Config.Validate()` then reported:

```
models[0] has no model name
```

This is a false positive: the wizard intentionally leaves the model name empty so the entry floats with the vendor's default model.

## Change

- `internal/config/config.go`: Allow the first model entry to have an empty `model` when `DefaultModel` is empty (the "floating default" case). Other entries must still have a model name.
- `internal/config/validate_test.go`: Add a passing case for the floating default and move the "missing model name" assertion to a non-first entry.

## Verification

- `go test ./internal/config/ -run TestConfigValidate -v` passes.
- `go test ./internal/tools/ -run TestConfigGuard -v` passes.
- `go test ./cmd/octo/ -run 'TestConfig|TestModelPicker|TestEnsureSender|TestDoctor' -v` passes.

### DIFF
--- a/cmd/octo/chat.go
+++ b/cmd/octo/chat.go
@@ -48,6 +48,7 @@ func soulMissing() bool {
 
 // offerOnboarding asks (on a first run) whether to run the onboarding ceremony.
 // Default is yes; "n"/"no"/"s"/"skip" declines. Returns true to run /onboard.
+// Kept for tests; the CLI/TUI now auto-starts onboarding instead of prompting.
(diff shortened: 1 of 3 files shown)


Leihb:
Superseded by #1465 — the wizard now writes the concrete default model name instead of leaving model empty.

PR #1606, opened 2026-07-19
blog: fix editing residue and parallel() example in deep-dive posts

Leihb:
## What

Small correctness pass over the architecture deep-dive posts (zh + en):

- **zh**: ungarble the "one registry, many species" sentence — it had orphan quoted fragments (`"浏览器: internal/browser""workflow: ..."`) that read as broken text; rewrite as a proper sentence
- **zh**: fix stray `Concurrent` left mid-sentence in the syncSem description (`信号量（syncSem）Concurrent 同步 ...` → `控制同步 ...`)
- **zh + en**: drop stray word `token` in the DuckDuckGo cooldown sentence (`a token goroutine` → `a stampede of goroutines`)
- **zh + en**: fix the workflow `parallel()` example — the real DSL is `parallel(items, &blk)` (see `internal/workflow/prelude.rb`), which takes items plus a block. The old example passed an array of lambdas with no block and would not run.

## Notes

The two factual errors from earlier drafts (the nonexistent `~/.octo/memory.sqlite (pgvector)` row and the `MaxConcurrent 默认无上限` claim) are already fixed on main — the memory section now correctly describes the pluggable external backends, and the concurrency cap is described as hardcoded 8. This PR only cleans up what remains.

No code changes; markdown only.


Leihb:
Superseded by #1607 — the deep-dive post is being fully replaced with the v2 rewrite, which removes the v1 paragraphs this PR patched.

PR #1659, opened 2026-07-21
test(browser): deflake windows-latest browser tests + fix download-attribution race

Leihb:
## Problem

PR #1657 (a landing-page-only change) failed the windows-latest Go job twice, on **different** subsets of `internal/browser` tests each time:

- Attempt 1: `TestRecorderCapturesActions`, `TestRecorderCapturesNewTab_ClickBeforeInstrument`, `TestAutoWaitNetwork`, `TestAutoWaitElement`, `TestSelectorSemanticAnchor` — all `wait for …: timed out after 20s`
- Attempt 2: `TestOOPIFReplay`, `TestOOPIFTypeReplay` — both `navigate: timed out waiting for load` (30s)

Same root cause both times: on a starved 2-core windows-latest runner, a cold headless Chrome intermittently needs more than 20-30s to finish its **first** load of a local fixture page. The random subset per run is the signature of runner starvation, not a code defect (main is green on the same code).

## Fix

Two-pronged: reduce the actual latency, then give the residual tail headroom.

**1. Quiet Chrome down (`testChromeArgs`)** — the test launch (shared by `newBrowser`, the OOPIF helper, and the digest test, which previously launched bare) now disables the background work that competes with the page under test on 2 cores: sandbox broker processes, background networking, component update, crashpad, sync, safe-browsing phishing classification, tab throttling, etc. Safe because every test drives a throwaway profile against local `httptest` fixtures. Launch has no non-test callers (attach-only since #1472), so production is untouched.

**2. Raise the ceilings that only bind when things are already slow:**

| ceiling | before | after |
|---|---|---|
| `testWaitTimeout` (WaitFor) | 20s | 60s (3rd bump: 5s flaked TestUpload, 20s flaked five tests on 2026-07-21) |
| `Navigate` load event | const 30s | package var, still 30s in production; tests raise to 90s |
| whole-test ctx on Chrome-launching tests | 30-60s | 180s, so the per-step ceilings above get to bind and name the slow step |
| CI `go test` | default 10m/pkg | `-timeout 20m`, so a bad-weather browser-package run fails cleanly instead of panicking |

All ceilings are free on healthy runs — WaitFor/Navigate return the instant the condition holds. Pure-mock tests (ws-url resolution etc.) keep their tight 5s/250ms budgets.

## Verification

`go test ./internal/browser/` green locally (real Chrome, 144s, macOS) — including the OOPIF site-isolation tests under the new flag set. `gofmt`/`go vet` clean.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -83,5 +83,10 @@ jobs:
       # catches platform-specific failures (path handling, process spawning, the
       # GOOS-guarded cases). The race pass on Windows was the CI bottleneck
       # (~150s of a ~220s job) thanks to instrumentation + costly process spawns.
+      # -timeout 20m: the browser package launches a real Chrome per test and
+      # already runs ~6-8min on a starved windows-latest runner; with the
+      # deliberately generous per-test ceilings (see internal/browser
+      # testWaitTimeout) a bad-weather run must fail those tests cleanly
+      # instead of hitting the go-test default 10m panic.
       - name: go test
-        run: go test ${{ matrix.os == 'ubuntu-latest' && '-race' || '' }} ./...
+        run: go test -timeout 20m ${{ matrix.os == 'ubuntu-latest' && '-race' || '' }} ./...
--- a/internal/browser/browser.go
+++ b/internal/browser/browser.go
@@ -424,10 +424,16 @@ func (p *Page) Navigate(ctx context.Context, url string) error {
 		return ctx.Err()
 	case <-events:
 		return nil
-	case <-time.After(30 * time.Second):
+	case <-time.After(navigateLoadTimeout):
 		return fmt.Errorf("navigate %s: timed out waiting for load", url)
 	}
 }
+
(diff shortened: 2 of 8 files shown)


Leihb:
Pushed a second commit covering a **third** flake mode that surfaced on #1657's next rerun: `TestAutoDownloadDetection` — the click was never upgraded to a download event.

Root cause is a genuine order-dependence in the recorder, not just slowness: `Browser.downloadWillBegin` is emitted by the browser process, while the click event rides the renderer → binding → CDP path. On a loaded machine the download event can arrive **before** the click it belongs to; `upgradeLastClickToDownload` found no click to upgrade and dropped the download irrecoverably.

The recorder now parks an unmatched `downloadWillBegin` and the next click within the attribution window claims it (order-independent matching). The 5s attribution window is now a var — production default unchanged, tests widen it to 30s — and the test polls up to `testWaitTimeout` instead of a flat 5s.

Full `internal/browser` suite green locally against real Chrome (142s).

PR #1707, opened 2026-07-22
fix(web): animate tool card auto-collapse instead of snapping shut

Leihb:
## Summary
- The last tool card in a `ToolGroup` auto-opens while its turn streams and auto-collapses once nothing is running; binding `<details open>` straight to that default made the collapse instant, which read as a UI flash/bug.
- The `<details>` now stays forced open for ~2s while a CSS `grid-template-rows` transition shrinks its content first, so the native collapse that lands afterward is invisible.
- A user-driven click on the summary still collapses natively/instantly — only the automatic, no-override default-driven collapse is animated.

## Test plan
- [x] `svelte-check` — 0 errors
- [x] `vitest run src/lib/toolFold.test.ts` — 9 passed (unchanged, confirms default-open logic untouched)
- [x] `make web-build`
- [ ] Manual: run a turn with several tool calls, confirm the previously-last card now shrinks smoothly over ~2s instead of vanishing instantly once the turn finishes

### DIFF
--- a/web/src/components/chat/ToolGroup.svelte
+++ b/web/src/components/chat/ToolGroup.svelte
@@ -3,7 +3,7 @@
   import { activeSessionId, showToast } from '../../lib/stores'
   import { ws } from '../../lib/ws'
   import * as api from '../../lib/api'
-  import { toolOpenState, applyToolToggle } from '../../lib/toolFold'
+  import { toolOpenState, applyToolToggle, defaultToolOpen } from '../../lib/toolFold'
 
   // Tracks which overwrite-undo buttons have already fired, keyed by tool id.
   let undone = $state<Record<string, boolean>>({})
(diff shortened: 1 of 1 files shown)


Leihb:
Follow-up commit narrows the scope: only the group-level running→not-running edge (the turn actually finishing) animates now. The earlier version also animated the mid-stream handoff (an older tool card losing its auto-open default the instant a *newer* tool call becomes the new last one) — that instant collapse always worked fine on its own, but animating it concurrently with the new card's arrival read as if the open was animating too. That handoff case is back to its original instant-collapse behavior; only the true end-of-turn collapse gets the 2s shrink.

Leihb:
Combined into #1708 — rebased this branch's commits underneath it per request, so this now duplicates that PR. Closing in favor of a single PR.

PR #1753, opened 2026-07-24
chore(subagent): drop the built-in plan preset

Leihb:
## Summary

Removes the built-in `plan` sub-agent preset (`explore` / `general` / `code-review` remain).

**Why it's dead weight:**

- It's a Claude Code carryover without its native habitat — CC's Plan agent exists to serve Plan Mode, which octo doesn't have.
- The `sub_agent` tool's main Description steers the model only toward explore / general / code-review; `plan` appears solely in the parameter enum, so the model is never nudged to pick it.
- Its delta over `explore` is a single persona paragraph (same `readOnly`, same `leanSystem`) — `explore` plus "then produce an implementation plan" in the prompt is equivalent.
- Nothing in the repo (skills, workflows, docs) ever dispatches it; the project's own planning path is the interactive grill-me / tech-design / implement skills.

**Safety:** `subagent_type` is a free-form string, so a model that still asks for `"plan"` gets a clear `unknown subagent_type "plan". Available: …` error listing the remaining presets and recovers. Anyone who wants it back can define `~/.octo/agents/plan.md` — user-defined agents take precedence over built-ins anyway.

**Also updated:** the `plan` mentions in the docs site (en + zh sub-agents guide), dev-docs, the `ComposePair` comment, and both architecture-deep-dive blog posts (which also carried the already-debunked "server/IM forces sync" claim from #1751 — fixed while touching those lines).

## Test plan

- `go build ./...`, `go vet`, `gofmt -l` — clean
- `go test ./internal/tools/ ./internal/prompt/` — pass
- `go test -race ./...` — green


Leihb:
Merged manually via local squash (GitHub API refused to merge with --admin due to a stale merge-conflict flag). Commit b2178412 on main.

PR #1754, opened 2026-07-24
feat: count grep tool results as reads for read-before-write

Leihb:
## 问题

`grep` 工具已经让模型看到了要改的那几行原文，但 ReadTracker 只给 `read_file`/`write_file`/`edit_file` 和 terminal 读命令盖章——所以 grep-then-edit 会撞 `File has not been read yet`，被迫对整个文件做一次冗余的 `read_file`（大文件尤其浪费 token）。

## 改动

`internal/tools/registry.go` 的 dispatch 成功分支新增 `case "grep"`：`recordGrepReads` 解析 grep 输出，给每个出现过的文件盖章——

- `files_with_matches` 模式：整行即路径
- `content` 模式：`path:N:text`（匹配行）、`path-N-text`（上下文行）
- `count` 模式：`path:N`
- 单文件搜索（rg 输出不带路径前缀）：取 input 的 `path` 参数

非路径行（`(no matches)`、`--` 分隔符、截断标记）stat 失败自然跳过。

## 安全性

- 盖章同时钉住 mtime，之后发生 out-of-band 修改时 "modified since read" 守卫照常触发
- grep 没有命中的文件仍然不可写（有测试覆盖）

## 测试

新增 4 个用例：目录 grep 后可编辑、files_with_matches 后可编辑、单文件 grep 后可编辑、未命中文件仍被拦截。`go test ./internal/tools/...` 全绿。

### DIFF
--- a/internal/tools/read_tracker_test.go
+++ b/internal/tools/read_tracker_test.go
@@ -509,3 +509,27 @@ func TestRegistry_GrepContextModeThenEdit_Allowed(t *testing.T) {
 		t.Errorf("edit after context-mode grep should succeed: %v", err)
 	}
 }
+
+// Filenames with embedded separator+digit runs (e.g. "report-2024-01-01.txt")
+// used to fail closed: the non-greedy regex settled on the first "-2024-"
+// boundary and the candidate "report" didn't stat. The multi-boundary fix
+// now tries every prefix, so the full filename is found.
+func TestRegistry_GrepNumericDashFilename_Allowed(t *testing.T) {
+	reg := NewDefaultRegistry()
+	dir := t.TempDir()
+	p := filepath.Join(dir, "report-2024-01-01.txt")
+	if err := os.WriteFile(p, []byte("line one\nconst a = 1\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := reg.Execute(context.Background(), "grep", map[string]any{
+		"pattern": "const a", "path": dir,
+	}); err != nil {
+		t.Fatalf("grep: %v", err)
+	}
+	if _, err := reg.Execute(context.Background(), "edit_file", map[string]any{
+		"path": p, "old_string": "const a = 1", "new_string": "const a = 2",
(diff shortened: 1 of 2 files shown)


Leihb:
## Self-review findings (addressed in 1887b9ee)

Ran an isolated code-review pass; it surfaced one Important issue and a few minor notes.

### Important — fixed
- **Zero-match single-file grep still stamped the input path.** `grep` with no matches returns `"(no matches)"` with a nil error, so the unconditional `record(input["path"])` stamped the file even though the model saw nothing — silently unlocking a blind edit. Fixed by skipping the input-path stamp when output is the no-match marker. Regression test added.

### Minor — noted, not changed
- Filenames with embedded `-<digits>-` (e.g. `report-2024-01-01.txt`) fail closed and don't get stamped. Safe direction (can't wrongly unlock), just a gap in "grep counts as read". Leaving as-is for now.
- Double stat per candidate (one in record, one inside RecordRead) — same pattern as recordTerminalReads, convention-consistent.

### Test gaps filled
- `TestRegistry_GrepSingleFileNoMatch_EditStillBlocked` — regression for the Important fix
- `TestRegistry_GrepCountModeThenEdit_Allowed` — count mode (`path:N`)
- `TestRegistry_GrepContextModeThenEdit_Allowed` — context mode (`path-N-text` + `--` separators)

Leihb:
## Minor #1 fixed (17de1c82)

**Filenames with `-<digits>-` (e.g. `report-2024-01-01.txt`) now get stamped correctly.**

Root cause: the non-greedy `^(.+?)[-:]\d+[-:]` settled on the first `-2024-` boundary, yielding candidate `report` which never stat-ted → file stayed unwritable.

Fix: replaced the single-prefix regex with `FindAllStringIndex` over `[-:]\d+[-:]`, trying every boundary as a prefix. The longest candidate that resolves to a real file wins.

Test: `TestRegistry_GrepNumericDashFilename_Allowed`.

Minor #2 (double stat) and #3 (naming) still outstanding — neither has functional impact.

PR #1776, opened 2026-07-25
fix(server): validate tool names in agent profile CRUD

Leihb:
## Summary

The REST API for creating/updating agent profiles accepted any string in the
`tools` field. Unknown tool names were silently ignored at runtime when the
profile's tool allowlist was evaluated — no feedback to the user.

## Changes

1. **internal/tools/registry.go**: Export `KnownToolNames()` returning all
   built-in tool names from the canonical `allTools` list.
2. **internal/server/agents_handlers.go**: Add `unknownToolNames()` helper
   that checks names against known tools + known skills. Call it in both
   `handleCreateAgent` and `handleUpdateAgent` before writing the profile.

## Test plan

- Existing `TestHandleAgents_*` all pass (none provide unknown tools).
- `go vet ./...` clean.
- Manual: POST `{"tools":["nonexistent_tool"], ...}` returns 400.

### DIFF
--- a/internal/server/agents_handlers.go
+++ b/internal/server/agents_handlers.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 
 	"github.com/open-octo/octo-agent/internal/agentprofile"
+	"github.com/open-octo/octo-agent/internal/skills"
+	"github.com/open-octo/octo-agent/internal/tools"
 )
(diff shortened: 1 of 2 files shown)


Leihb:
Rebased into #1775

PR #1841, opened 2026-07-26
fix(glob): handle absolute-path patterns correctly

Leihb:
When the glob pattern starts with an absolute path (e.g. /Users/.../src/**/*.go), the tool now uses the pattern's literal directory prefix as the search root instead of the current working directory.

Two-part fix:
1. literalPathPrefix: preserve the leading '/' when pattern is absolute, so the prefix resolves as an absolute path.
2. Execute: when pattern is absolute, extract its literal prefix as absRoot and strip it from the pattern before matching. All downstream logic (scan-root pruning, file enumeration, globMatch) then operates naturally on the adjusted absRoot + matchPattern.

Also fix i18n: change '通过智能体创建' / 'Create with Agent' → '创建智能体' / 'Create Agent' on the Agents panel button.

## User-visible impact

---

## Self-review

### 1. Architecture

### 2. Need

### 3. Code

- [ ] All tests pass

### Bonus

---


Leihb:
Superseded by main: the absolute-path glob handling, tests, and i18n changes are already present in the current main branch.

PR #1846, opened 2026-07-27
feat(upgrade): call out a running serve daemon on upgrade failure and success

Leihb:
## Summary

Follow-up to #1842 (item 3 of the planned improvements). The reporter ran `octo upgrade` on Windows while `octo serve` was running and the binary swap "silently" failed — a running executable is locked on Windows, and even when the swap succeeds the daemon keeps executing the old image.

`octo upgrade` now detects a live daemon through the shared `~/.octo/serve.pid` contract (`runningServeDaemon`, reusing the `serveproc` aliases from `serve_daemon.go`) and adjusts its messaging:

- **On failure**: appends a hint that the daemon (pid shown) may be holding the binary in place, with the recovery sequence — `octo serve stop`, retry the upgrade, `octo serve -d`.
- **On success**: states explicitly that the running daemon still uses the old binary and how to restart it, replacing the generic "picks the new binary up on its next restart" note.

Detection is best-effort guidance, never a gate: the upgrade behavior itself is unchanged, and a foreground `octo serve` (no pid file) stays invisible.

## Testing

- `TestRunningServeDaemon` — pid-file detection (isolated via the swapped `daemonPidFile`, faking a live daemon with the test process's own pid).
- `TestRunUpgrade_FailureHintsRunningServeDaemon` — full `runUpgrade --force` against an httptest origin that serves only `releases/latest` (asset download 404s), mirrors cleared so nothing leaves the fake server; asserts the hint lands on stderr.
- `go test -race ./cmd/octo/`, `go vet`, `gofmt -l` all clean.

Closes the remaining code change promised in #1842.

### DIFF
--- a/cmd/octo/upgrade.go
+++ b/cmd/octo/upgrade.go
@@ -43,35 +43,38 @@ func runUpgrade(args []string, stdout, stderr io.Writer) int {
 		return 0
 	}
 
-	daemonPid, daemonRunning := runningServeDaemon()
-
 	err := upgrade.Run(ctx, upgrade.Options{
 		Force: *force,
 		Log:   func(line string) { fmt.Fprintln(stdout, line) },
 	})
+	// Checked after Run, not before: a download can take minutes, and what
+	// matters is who is alive when the message prints.
+	backendPid, backendRunning := runningServeDaemon()
 	switch {
 	case errors.Is(err, upgrade.ErrUpToDate):
 		fmt.Fprintln(stdout, "already up to date")
 		return 0
 	case err != nil:
 		fmt.Fprintf(stderr, "octo upgrade: %v\n", err)
-		if daemonRunning {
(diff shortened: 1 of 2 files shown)


Leihb:
Review follow-ups pushed in 2af3c95:

- **Hedged both messages** — `~/.octo/serve.pid` is a shared contract and may belong to the desktop hub (a different executable that never locks the CLI binary), so the daemon claim and the `octo serve stop` advice are now phrased conditionally ("if it's an `octo serve -d` daemon …").
- **Backend snapshot moved after `upgrade.Run`** — downloads can take minutes; liveness at message-print time is what matters for both the lock hint and the restart note.
- Restart advice now says "with your usual flags" so `--addr` etc. aren't silently dropped.
- Pinned the negative case: a failure with no backend registered prints no hint.

Known gap (deliberate): the success-path message variant is untested — `runUpgrade` doesn't expose `upgrade.Options.TargetPath`, so a successful in-test `Run` would swap the actual test binary; adding a seam just for this message wasn't worth it.

PR #1889, opened 2026-07-29
fix(web): render image artifacts outside the sandboxed preview iframe

Leihb:
Supersedes #1888, which fixed the same symptom by base64-ing the image into the `srcdoc` document. This takes the iframe out of the picture for images instead.

## Root cause

An `` placed inside the Artifacts panel's `srcdoc` iframe cannot authenticate. The iframe carries no `allow-same-origin`, so its subresource request has an opaque origin and counts as cross-site; the `SameSite=Strict` `octo_access_key` cookie is withheld and the endpoint answers 401.

Loopback clients are exempt from auth altogether (`requireAuth`'s hardened loopback path), which is why this never reproduced on `localhost` — it only breaks when the UI is reached over a tunnel, the LAN, or `--host`. It is not size-dependent; every image artifact was broken for remote clients.

The response's `Content-Security-Policy: sandbox` header is not involved: the `sandbox` directive is only enforced when a response is loaded as a document, not for an `` subresource.

## Change

Images render as a plain `` in the host document, whose own request is same-site and authenticates normally. `Artifact` gains a `src` field for that URL; the three render sites (desktop panel, maximized modal, mobile viewer) branch on it.

Side benefits:

- **Lazy loading is back.** History replay no longer needs the bytes at all, so a session full of multi-MB browser screenshots costs nothing until the user opens one. The base64 approach downloaded every image on replay, concurrently, and kept each one resident as a UTF-16 string.
- **Replay keeps transcript order** for images, since the image branch no longer awaits.

Dropping the iframe costs no isolation here: the endpoint pins `Content-Type` per extension and sends `X-Content-Type-Options: nosniff`, and an SVG loaded through `` runs no script and fetches no external resource.

Two consequences of the split:

- An image has no source view, so the Preview/Code toggle is hidden (the mobile viewer already did this) and `code` carries the on-disk path — the one text form worth copying.
- **Download now saves the actual bytes.** Previously `code` held the fetch URL, so downloading a `.png` wrote a text file containing a URL. Desktop routes through `nativeSaveBinary`, same shape as the skill zip export.

The file-header comment in `lib/artifacts.ts` now states the constraint, because it applies to every preview document built there, and `artifacts.test.ts` pins both halves: images stay out of the iframe and fetch nothing on observe; generated preview chrome contains no `/api/` reference.

## Known remaining gap

An image reference written by hand inside a **markdown** artifact (`shot`, or an absolute `/api/…` path) still fails inside the iframe for the same reason — that content is passed through untouched. Out of scope here; worth a follow-up issue.

## Verification

- `vitest run` — 61 passed (5 files, 3 new)
- `svelte-check` — 0 errors (1 pre-existing warning in `LightAppsView.svelte`)
- `vite build` — 298 modules, ok
- No Go changes.

Not verified by me: the remote repro itself. Worth a pass over ngrok on an image artifact, an SVG, and the download button on both desktop and browser.

### DIFF
--- a/web/src/components/ArtifactModal.svelte
+++ b/web/src/components/ArtifactModal.svelte
@@ -13,7 +13,7 @@
   })
 
   function onCopy() { copyArtifact(cur?.code ?? '', showToast) }
-  function onDownload() { downloadArtifact(cur?.name, cur?.code ?? '', showToast) }
+  function onDownload() { downloadArtifact(cur, showToast) }
 
   // "Back to sidebar": close modal, reopen the side panel.
   function restoreSidebar() {
@@ -53,7 +53,12 @@
 
     <!-- Body — always preview, no toolbar / footer chrome -->
     <div class="body">
-      <iframe srcdoc={cur.preview} sandbox="allow-scripts clipboard-write" title={cur.name}></iframe>
+      {#if cur.src}
+        <!-- Images render outside the sandboxed iframe (see lib/artifacts.ts). -->
+        <div class="img-wrap"><img src={cur.src} alt={cur.name} /></div>
+      {:else}
+        <iframe srcdoc={cur.preview} sandbox="allow-scripts clipboard-write" title={cur.name}></iframe>
+      {/if}
     </div>
   </div>
 </div>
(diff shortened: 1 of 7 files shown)


Leihb:
Superseded — this landed on #1888's branch instead, plus the markdown-image half of the same bug.

PR #1918, opened 2026-07-30
fix(web): keep the composer draft when picking a slash command

Leihb:
## Problem

Reported from the web UI: type a message, then click the "/" button to pick a skill — the whole draft disappears, leaving just `/` in the composer.

`insertSkill()` did `text = '/'` unconditionally, and `selectItem()` then overwrote `text` with the picked command.

## Fix

A slash command is a whole-message prefix, so an existing draft is naturally its **argument**, not something to discard:

- `insertSkill()` leaves the textarea alone when the composer is non-empty; it parks the draft in `slashDraftTail` and opens the menu. An empty composer still gets `/` prefilled so typing filters the list.
- A half-typed `/query` counts as the trigger, not a draft (no `/code-review /`).
- `selectItem()` builds the command string and joins the parked draft onto it via `composeSlashCommand()`.
- `hideSlashMenu()` clears the parked draft, so Escape / click-outside / typing away leaves the draft exactly as the user wrote it.

Example: draft `check the diff` + pick `/code-review` → `/code-review check the diff`.

## Tests

- New `web/src/lib/slashCompose.ts` + unit tests for the join rule.
- `npm run test` — 8 files, 92 tests pass.
- `svelte-check` — 0 errors.

### DIFF
--- a/web/src/components/chat/Composer.svelte
+++ b/web/src/components/chat/Composer.svelte
@@ -10,6 +10,7 @@
   import * as api from '../../lib/api'
   import { t } from '../../lib/i18n'
   import { submitIntent } from '../../lib/composerKeys'
+  import { composeSlashCommand } from '../../lib/slashCompose'
   import StatusTag from '../ui/StatusTag.svelte'
   import FolderPickerModal from '../overlays/FolderPickerModal.svelte'
   import type { McpServerDetail, McpTool } from '../../lib/types'
@@ -296,6 +297,10 @@
   let slashQuery = $state('')
   let slashActiveIndex = $state(-1)
   let slashMcpServer = $state('')
+  // Draft carried across the menu when the "/" button was pressed on a
+  // non-empty composer; the picked command is prefixed onto it. See
+  // insertSkill/composeSlashCommand.
+  let slashDraftTail = $state('')
 
   type SlashItem =
     | { kind: 'builtin'; name: string; descKey: string }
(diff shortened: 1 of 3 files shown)


Leihb:
Second commit — follow-ups from an isolated code review of the first one.

The parked draft was only ever released by `hideSlashMenu()`, and three paths never reach it:

- **Stale draft resurrected.** Park a draft via the "/" button, then select-all and retype `/` — `handleSlashInput` calls `showSlashMenu` directly, so the old draft survived and got appended to the command the user then picked. `showSlashMenu()` now drops it; `insertSkill` parks after that call.
- **Menu outliving its message.** `slashActiveIndex` starts at -1 and hovering doesn't set it, so Enter falls through to `send()`, which cleared the text but left the menu open — a later click composed the command onto the message already sent. `send()` now closes it.
- **Cross-session leak.** ⌘K session switching fires no outside-click, so both the menu and the draft rode into the next conversation. The per-session draft `$effect` now closes it.

Two more:

- `/clear` and `/compact` are matched by exact equality server-side (`handleSlashCommand`), so appending a draft silently downgraded them to an ordinary chat message. `BUILTIN_COMMANDS` now carries `takesArgs`, and those two ignore a parked draft.
- The workflow / MCP-tool prefills are whole sentences, not `/name ` prefixes; the draft now lands on its own line instead of running into the end of one (`"…what params are needed check the diff"`).

Not changed: with a non-empty composer the "/" button opens the menu without touching the textarea, so type-to-filter isn't available on that path (typing dismisses the menu, draft intact). Prefilling `/` to keep filtering would hide the draft while the menu is open and needs a restore-on-dismiss rule that fights with typing-away — not worth it for a mouse-driven affordance. Filtering still works the normal way: type `/` into an empty composer.

`npm run test` 92 pass, `svelte-check` 0 errors.

PR #1929, opened 2026-07-30
fix(web): order artifacts by transcript position, not fetch completion

### LINKED ISSUE #1894: Artifacts list order and auto-selection follow fetch completion, not the transcript
The artifacts list ends up in fetch-completion order, not the order the files appear in the transcript, and the auto-selected artifact after a history replay is whichever fetch finished last.

`observeArtifact` is called fire-and-forget (`ChatView.svelte:313`, `mobile/chatWiring.ts:66`), awaits a fetch, then does:

```js
artifacts.update(...)                       // append
artifactSel.set(get(artifacts).length - 1)  // select the newest
```

So every artifact selects itself as it lands, and the last one to *resolve* wins — not the last one in the transcript.

## Observed

With a transcript of `[report.md, late.png]`, the footer chips come out as `[late.png, report.md]`.

Reopening a session can therefore land on an arbitrary artifact rather than the most recent one. #1888 widens the gap: a markdown or HTML file with N local images now needs 1+N sequential round-trips before it lands, so it reliably loses the race to smaller artifacts written after it.

## Notes toward a fix

Replay knows the transcript order — the payloads arrive in it. Ordering the store by that index rather than by arrival, and choosing the selection from the same index rather than `length - 1`, would fix both symptoms. Live turns can keep appending at the end.

Pre-existing (text artifacts have always awaited a fetch); surfaced by the review of #1888.

Leihb:
Fixes #1894.

## Problem

`observeArtifact` awaits a per-artifact fetch and then appends to the `artifacts` store and selects `length - 1`, so the footer chips came out in fetch-completion order and the auto-selected artifact after a history replay was whichever fetch resolved last — not the last artifact in the transcript. #1888 widened the gap: a markdown/HTML file inlining N local images needs 1+N sequential round-trips, so it reliably lost the race to smaller artifacts written after it.

## Fix

All contained in `web/src/lib/artifacts.ts` (plus one field on the `Artifact` type):

- Assign a monotonically increasing `seq` **synchronously, before the fetch starts**. Both replay loops (`ChatView.svelte`, `mobile/chatWiring.ts`) and the live `tool_result` handlers call `observeArtifact` in transcript order, so `seq` captures transcript order without any caller changes.
- Upsert sorted by `seq` instead of appending. The list is then invariantly in transcript order, so the existing `length - 1` selection now lands on the transcript-newest artifact regardless of which fetch finished last.
- A re-observed path gets a new `seq` and moves to its new position (same as the old move-to-end behaviour); a stale observation of a path whose later observation already landed is dropped instead of displacing it.
- `resetArtifacts` resets the counter along with the rest of the per-session state.

Live turns keep the previous behaviour: artifacts append at the end and the newest is selected.

## Tests

Two new cases in `artifacts.test.ts`, including a direct reproduction of the issue scenario — `[report.md, late.png]` where the markdown fetch is held open until after the image lands — asserting both the final order and the selection. `svelte-check`: 0 errors; full web suite: 108 passed; `vite build` clean.

### DIFF
--- a/web/src/lib/artifacts.test.ts
+++ b/web/src/lib/artifacts.test.ts
@@ -375,5 +375,7 @@ describe('observeArtifact — transcript ordering', () => {
 
     expect(get(artifacts)).toHaveLength(1)
     expect(get(artifacts)[0].code).toBe('v2')
+    // The dropped stale observation must not touch the selection either.
+    expect(get(artifactSel)).toBe(0)
   })
 })
--- a/web/src/lib/artifacts.test.ts
+++ b/web/src/lib/artifacts.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { get } from 'svelte/store'
-import { artifacts } from './stores'
+import { artifacts, artifactSel } from './stores'
 import { observeArtifact, resetArtifacts } from './artifacts'
 
 // Nothing a preview document references can authenticate: the srcdoc iframe has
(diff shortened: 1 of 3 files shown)


Leihb:
Reviewer note: the `artifactSelSession` guard is keyed on session id equality, so a same-session `history_reload` (after /clear or an edit truncation) does not fence off the previous generation's in-flight fetches — a pre-existing weakness this PR neither fixes nor worsens. If it ever bites, replacing the session-id marker with a generation token bumped in `resetArtifacts` would close it.

PR #1931, opened 2026-07-30
fix(web): save the inlined preview to Light Apps so local images survive

### LINKED ISSUE #1890: Save to Light App stores an HTML artifact's raw source, so its local images break
Split out of #1888, which introduced the asymmetry.

## What happens

An HTML artifact that references its own local images — ``, `background-image:url(bg.png)` — now previews correctly in the Artifacts panel: #1888 inlines those bytes as `data:` URIs, because nothing inside the sandboxed `srcdoc` iframe can authenticate against `/api/*`.

"Save to Light App" passes `cur.code`, the raw source (`web/src/components/ArtifactsPanel.svelte:30`):

```js
const app = await api.createLightApp({ name, html: cur.code })
```

A Light App renders through the same kind of sandboxed iframe (`ArtifactsPanel.svelte:110`), and its relative paths have nothing to resolve against at all. So the user sees the images fine in the panel, clicks Save to Light App, opens it, and every image is broken.

Before #1888 both were broken — consistently. The new asymmetry is the problem: it reads as a Light Apps bug.

## Why it wasn't fixed in #1888

Neither option is obviously right:

- **Store the inlined preview.** The Light App becomes self-contained and actually works. But it permanently persists the base64 in the Light App store — megabytes per screenshot, for a copy that can never be re-derived from the original file.
- **Keep storing the source.** No storage bloat, and the Light App stays a faithful copy of what the agent wrote. But it is knowingly broken for any artifact with local images, and the panel now sets the opposite expectation.
- **Refuse the save**, or warn, when the artifact has local image references — cheap, honest, and unhelpful.

A fourth option: give Light Apps their own asset storage so relative references resolve. That's a real feature, not a bug fix.

## Reproduce

1. Have the agent write an HTML file and a PNG beside it, with the HTML referencing the PNG by relative path.
2. Open the Artifacts panel — the image renders.
3. Save to Light App, then open it from the Light Apps panel — the image is broken.

Worth noting this is only visible over a tunnel/LAN for the *panel* half; on loopback the panel worked before #1888 too, because loopback clients skip auth.

Leihb:
Fixes #1890 (split out of #1888, which introduced the asymmetry).

## What

"Save to Light App" passed `cur.code` — the artifact's raw source — to `createLightApp`. A Light App renders through the same kind of sandboxed `srcdoc` iframe as the panel preview, but its relative image references have nothing to resolve against at all, so every `` / `url(bg.png)` was broken in the saved copy while the panel preview (which #1888 fixed by inlining those bytes as `data:` URIs) showed them fine.

## How

Persist what the panel previews, not the raw source. A new `lightAppSource(artifact)` helper in `web/src/lib/artifacts.ts` picks the version to save:

- **HTML with local images** → the inlined preview (`data:` URIs), so the Light App is self-contained and renders the way the panel did.
- **HTML routed to the warning page by `hasExternalRefs`** → the raw source. The preview is a placeholder there, not the document; persisting it would be strictly worse than the broken images.
- **Nothing to inline** → the raw source (the preview is byte-identical to it in that case, per `inlineLocalRefs`'s no-op path).

This is the "store the inlined preview" option from the issue: the saved copy costs base64 bytes in `~/.octo/light-apps/` (bounded by the existing 4 MiB / 20-ref inlining budget), in exchange for a Light App that actually works. References beyond that budget stay relative and remain broken, same as the panel preview.

## Tests

New `lightAppSource` cases in `web/src/lib/artifacts.test.ts` pin all three branches above (built through `observeArtifact`, so they exercise the real preview pipeline). `vitest` 109/109 in `web/`, `svelte-check` 0 errors, `npm run build` clean.

### DIFF
--- a/web/src/components/ArtifactsPanel.svelte
+++ b/web/src/components/ArtifactsPanel.svelte
@@ -2,6 +2,7 @@
   import { artifacts, panelContent, artifactSel, artifactView, artifactModalOpen, lightappSel, lightapps, lightappHTML, showToast } from '../lib/stores'
   import { t } from '../lib/i18n'
   import { copyArtifact, downloadArtifact, imagePreviewError } from '../lib/artifact-actions'
+  import { lightAppSource } from '../lib/artifacts'
   import * as api from '../lib/api'
 
   // ── Session artifacts (existing) ──────────────────────────────────────────
@@ -51,7 +52,10 @@
     if (!name || !cur) return
     saveToLALoading = true
     try {
-      const app = await api.createLightApp({ name, html: cur.code })
+      // Save what the panel previews, not the raw source: a Light App renders
+      // in the same kind of sandboxed iframe, where relative image paths
+      // resolve against nothing (#1890).
+      const app = await api.createLightApp({ name, html: lightAppSource(cur) })
       showToast(`Light App "${app.name}" saved`, 'success')
       saveToLADialog = false
       // If the panel is in lightapps mode, refresh the list.
--- a/web/src/lib/artifacts.test.ts
+++ b/web/src/lib/artifacts.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { get } from 'svelte/store'
 import { artifacts } from './stores'
-import { observeArtifact, resetArtifacts } from './artifacts'
+import { lightAppSource, observeArtifact, resetArtifacts } from './artifacts'
 
 // Nothing a preview document references can authenticate: the srcdoc iframe has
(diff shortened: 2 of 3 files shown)


Leihb:
Follow-up commit: raised the inline budget 4 MiB → 6 MiB and the ref cap 20 → 40. Now that the inlined preview is what the save persists, the budget decides how much of a page survives it. 6 MiB is the practical ceiling under the `POST /api/light-apps` 10 MB body cap (raw bytes × ~1.37 as base64 ≈ 8.2 MB, leaving room for the document itself) — going past ~7 MiB would 413 the save without a matching server-side change.

Issue #1868, opened 2026-07-27
Session export: add PDF / PNG / JSON / HTML formats + message selection (Web UI)


Leihb:
> 中文读者：这是一个适合新人上手的任务。**改动全在 Web 前端，Go 后端零改动**，
> 设计文档（`dev-docs/session-export-enhancement.md`）已经把交互流程、状态管理、
> 每种格式的生成方式、i18n key、测试清单全部写好了，照着实现即可。
> 想认领的在下面留言说一声，我会回复确认，避免两个人撞车。

## Problem

Session export in the Web UI is currently minimal (`ChatView.svelte`, `exportTranscript()`):

- **Markdown only** — no PDF, PNG, JSON or HTML.
- **All-or-nothing** — you cannot export a subset of the conversation.
- **Drops tool and thinking content**, so the export is lossy.

If you want to share one specific exchange out of a long session, there is no way to do it today.

## Goal

1. Support five export formats, all generated locally in the browser and downloaded: **Markdown, PDF, PNG, JSON, HTML**.
2. Let the user **pick which messages** to export instead of always exporting everything.
3. Do it **in place** — clicking export enters an "export mode" with a sticky format bar, no modal dialog.

## Design doc

`dev-docs/session-export-enhancement.md` is the spec. It already covers:

- The export-mode state (`web/src/lib/exportStore.ts`), enter/exit paths (button, ESC, loading state)
- Per-message checkboxes in the `#each msgs` loop (only `user` / `assistant` messages are selectable; `tool` messages stay out of the selection but are included in JSON)
- The sticky top bar layout and the message counter
- Generation logic for each of the five formats
- The self-contained HTML template + DOMPurify sanitising
- New i18n keys (EN + ZH), test plan, compatibility notes

Please read it before starting. If you disagree with anything in it, say so in this issue — the doc is not sacred, but let's settle the design before code.

## Scope

**In scope** — Web UI only:

- `web/src/views/ChatView.svelte`
- `web/src/lib/exportStore.ts` (new)
- `web/src/lib/i18n.ts`

**Out of scope**: desktop (`nativeShell`) export path, IM and TUI export. The desktop branch temporarily falls back to the browser download path; a follow-up can restore native save.

**Backend: no changes.** `GET /api/sessions/{id}/messages` already returns the full events array; everything happens client-side.

## New dependencies

The doc calls for three frontend packages: `jspdf`, `html2canvas`, `dompurify` (+ `@types/dompurify`).

[...]

PR #2002, opened 2026-08-03
feat(web): add session export mode with Markdown and JSON formats

eternalweightlessness:
Replace the single-click Markdown-only export with an in-place export mode that supports message selection and two formats.

- New web/src/lib/exportStore.ts: reactive stores for export-mode state and per-session message selection
- Sticky format bar with MD / JSON buttons, selected-count label, tool-include toggle, and cancel button
- Per-message checkboxes on user/assistant messages (default all selected; tool messages are not selectable)
- exportAsMarkdown() respects checkbox selection + optional tool-call inclusion via toggle
- exportAsJSON() fetches full server transcript for lossless backup
- ESC key exits export mode (merged into existing svelte:window)
- 9 new i18n keys in both EN and ZH
- No backend changes — all logic runs client-side

Related: #1868

## User-visible impact

---

## Self-review

### 1. Architecture

### 2. Need

### 3. Code

- [ ] All tests pass

### Bonus

---


Leihb:
## Review

Overview: replaces the single-click Markdown export with an export mode (sticky format bar, per-message checkboxes, MD/JSON formats, tool-include toggle). Pure frontend change, no backend touched. The interaction design is reasonable and the core store/UI code is clear, but a few things should be addressed before merge.

### Blocking

**1. Unrelated dependency additions + lockfile churn**
`web/package.json` adds `@emnapi/core` and `@rolldown/binding-win32-x64-msvc` to `dependencies`, but neither is referenced anywhere in `web/src`. The accompanying 637-line `web/package-lock.json` diff also strips `resolved`/`integrity` from many unrelated packages — this looks like a lockfile regenerated in a different environment/npm version/registry mirror and committed by accident. `@rolldown/binding-win32-x64-msvc` in particular is a Windows-only native binding and shouldn't be a direct project dependency. This violates the repo's "no new dependencies" / "smallest possible diff" guidance — please revert the unrelated parts of these two files.

**2. Unrelated `.gitignore` changes**
Adds `/.cursor/`, `/.windsurf/`, `/.reasonix/`, OS files, logs, `*.tsbuildinfo` — none related to session export. Looks like local tooling config that leaked in; please drop it from this PR.

### Worth confirming (may be intentional, but undocumented)

**3. Markdown export now reads from the local `chatMessages` store instead of the server transcript** (`selectedLocalMsgs()` / `exportAsMarkdown` in `ChatView.svelte`)
The removed comment explicitly justified fetching from the server: *"Fetch complete history from server instead of relying on the local chatMessages store, which may only contain a partial view of the session."* To support the new checkbox UI, the Markdown path now reads purely from local state, while the JSON path (`fetchEvents()`) still fetches the full transcript from the server. This means the two formats can diverge for the same session in edge cases where local and server state differ. Was this tradeoff intentional? If so, worth a note in the PR description; if not, the MD path should filter server-fetched data by selection instead.

**4. Checkbox selection has no effect on JSON export**
Once export mode is entered, all messages get checkboxes and the bar shows "N selected" — but `exportByFormat('json')` calls `fetchEvents()`, which dumps the full raw server transcript regardless of `selectedIds`. That matches the PR description's intent ("lossless full backup"), but there's no UI signal that switching to JSON ignores the selection — a user who deselects specific messages expecting them excluded, then exports JSON, will still get everything. Consider disabling/graying the checkboxes (or a tooltip) when JSON is the target format.

### Minor

- `exportStore.ts`: `isActive()` is defined but never called (`ChatView.svelte` reads `$exportModeStore[...]` directly instead); the destructured `set` in both store factories is also unused. Worth trimming so the API surface doesn't mislead future readers.
- The PR description says "9 new i18n keys" but 8 were added, and 3 of them (`chat.export_pdf`, `chat.export_png`, `chat.export_html`) have no corresponding export path in this PR (only MD/JSON buttons exist). Either drop the unused keys or note they're reserved for a follow-up.
- No test added for the new `web/src/lib/exportStore.ts`, even though sibling `lib/*.ts` modules in this repo generally have a colocated `*.test.ts` (e.g. `pendingSendOrder.test.ts`). Worth adding given it's pure store logic.
- The self-review checklist in the PR description is entirely unchecked.

PR #2013, opened 2026-08-04
fix(server): freeze the composed system prompt once per session, like the CLI

Leihb:
## Summary
- Web (`buildAgent`) and IM (`runChannelTurns`) sessions recomposed the entire system prompt from scratch on every turn — unlike the CLI/TUI, which compose once per process and freeze (`internal/prompt/prompt.go`'s own doc comment says as much). Two of the recomposed layers actually change mid-session in normal use: the live memory file (an agent's own tools write to it) and the env layer's per-day date line. Either changing invalidates the provider's prompt-cache prefix for that turn and every one after.
- `agent.Session` gains `ComposedSystem`/`ComposedLeanSystem`, persisted with the same append-or-rewrite mechanics as `Title`/`WorkingDir`/`PermissionMode` (`SetComposedSystem`).
- `buildAgent` and `runChannelTurns` now check the session's frozen fields first; only a session's first turn pays for `ComposePair` (and the memory/skills/MCP inputs that feed it) and persists the result.

## Trade-off (confirmed with the user before implementing)
A profile's system-prompt edit, or a skill/MCP toggle, now only takes effect in a **new** session — not one already open — same as editing those files under a running CLI process. Previously the server deliberately recomposed every turn so such edits landed immediately; this trades that immediacy for prompt-cache stability.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test -race ./...` (full suite green)
- [x] Updated `TestHandleChannelMessage_RefreshesSystemPerTurn` → `TestHandleChannelMessage_FreezesSystemAfterFirstTurn`, which asserted the old (now intentionally reversed) per-turn-recompose behavior

### DIFF
--- a/dev-docs/im-parity-design.md
+++ b/dev-docs/im-parity-design.md
@@ -32,15 +32,9 @@ config changes), every restart wiped every conversation.
   crash costs at most one turn, not a whole chain. Failures are logged to
   the server console and never eat the reply the user already received.
 
-## Per-turn memory freshness
+## Memory hooks parity
 
-`handleChannelMessage` recomposes the agent's system prompt (memory
-injection and the skills manifest included) at the start of every turn. Web
-turns always had this — `buildAgent` runs per turn — but the IM factory
-composed the prompt once at session creation, so memory written and skills
-imported mid-session were invisible to IM until a restart.
-
-IM turns also carry the L2 memory hooks, the same pair `buildAgent` wires
+IM turns carry the L2 memory hooks, the same pair `buildAgent` wires
 for web sessions: `UserInputHook` keyword reminders and the
 `ToolResultHook` save-nudge. The injector comes from the shared
 `injectorFor` store (keyed `im:<session key>`), is session-sticky — the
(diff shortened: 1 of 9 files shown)


Leihb:
## CodeQL alerts #169/#170 dismissed as false positives

CI flagged `go/clear-text-logging` (high severity ×2) on the new `slog.Warn("freeze composed system prompt", "session", sess.ID, "err", err)` line in `buildAgent` (server.go:1312). Pulled the full SARIF `codeFlows` for both alerts to verify before dismissing — full path (91–114 steps each):

```
internal/provider/anthropic/client.go:103  apiKey field → struct literal (Client)
  → internal/app/sender.go: client/sender construction
  → internal/server/server.go:1789-1798  sender assigned onto s.sender (*Server field)
  → internal/server/loop.go:161          wakerFor(sessionID) returns serverWaker{s: s, sessionID: sessionID}
                                          — embeds the WHOLE *Server (holding the API-key-carrying sender)
  → internal/tools/schedule_wakeup.go:39-40  WithWaker → context.WithValue(ctx, ctxKeyWaker{}, w)
  → internal/server/ws_handlers.go:1424  runCtx = tools.WithWaker(runCtx, s.wakerFor(sess.ID))
     (on the SAME ctx chain, a separate line stores sess.ID under a DIFFERENT key: ctxKeySessionID{})
  → internal/server/handlers_prepare_toolturn.go, bg_tasks.go, session.go: sessionID threads through
    to Session.ID / LoadSession's id fallback
  → internal/server/server.go:1312  sess.ID / err reach the new slog.Warn call
```

The actual API key string never flows into `sess.ID` or `err` — `err` is just a local file-write error from `Session.SetComposedSystem`, and `sess.ID` is a session identifier. The alert exists because CodeQL's `context.Value()` modeling doesn't disambiguate by key: once *anything* tainted is stored on a context chain (here, the whole `*Server` via `serverWaker`, under `ctxKeyWaker{}`), it treats *any* later read from that same chain — including `sess.ID` stored under the unrelated `ctxKeySessionID{}` key — as potentially returning the tainted value.

This is a pre-existing pattern (`wakerFor` embedding `*Server` in a context value) that this PR's new log line happened to be the first thing to surface it for — not a new leak this change introduced. An actual fix would mean not wrapping the whole `*Server` in `serverWaker`/context, which is a separate, broader refactor out of scope for this PR.

Dismissed both alerts as `false positive` with a shorter version of this reasoning (280-char API limit on the dismissal comment itself).

PR #2041, opened 2026-08-05
fix(web): route the JSON transcript export through the native save dialog

Leihb:
## Problem

In the desktop shell, exporting a session as **JSON** produced no file at all — the export bar closed as if it had worked. **Markdown** exported fine from the same bar.

`exportAsMarkdown` branched on `nativeShell` and handed the bytes to the OS save dialog. `exportAsJSON` always went straight to the in-page blob download, and the octo-served webview has no download delegate — so the `<a download>` click is a silent no-op there. `internal/server/native_handlers.go` already documents exactly this, which is why the native route exists in the first place.

A second defect compounded it: `exportAsJSON` returned `void`, so `ok` stayed `true` in `exportByFormat` and `exitExportMode()` ran — export mode closed and the user's checkbox selection was discarded with no error toast, making a failed export indistinguishable from a successful one.

## Fix

Both formats now go through one `deliverExport(content, filename, mime)` helper that owns the native-vs-blob decision and returns whether the save actually landed. The duplication between the two format functions was what let them drift apart, so collapsing it is the fix rather than a second copy of the branch.

## Verification

Exercised the real UI in Chrome via CDP against an isolated `HOME` with a seeded transcript, loading the page as `?shell=octo-desktop` and stubbing `/api/version` to report `native:true` so the frontend takes its desktop-shell path, with `/api/native/save-file` intercepted to observe the call.

Before, on `main`:

```
--- clicking MD    -> NATIVE SAVE CALLED, name = _.md,  content bytes = 89
--- clicking JSON  -> BLOB DOWNLOAD _.json          (a no-op in the real webview)
native save calls total: 1
```

After:

```
--- clicking MD    -> NATIVE SAVE CALLED, name = _.md,   content bytes = 89
--- clicking JSON  -> NATIVE SAVE CALLED, name = _.json, content bytes = 416
native save calls total: 2
```

Plain-browser export was also re-checked end to end and is unchanged: both formats still download as blobs, including a 2.3 MB payload.

`npx vitest run` — 144 passed. `npx svelte-check --threshold error` — 0 errors.

## Known adjacent issue, not fixed here

`title.replace(/[^\w.-]+/g, '_')` strips CJK, so a session titled 导出测试会话 saves as `_.json` / `_.md`. It predates this change and affects both formats equally; kept out to keep this one concept.

### DIFF
--- a/web/src/lib/filename.test.ts
+++ b/web/src/lib/filename.test.ts
@@ -1,0 +1,67 @@
+import { describe, it, expect } from 'vitest'
+import { filenameStem } from './filename'
+
+describe('filenameStem', () => {
+  it('keeps a CJK title intact', () => {
+    // The regression this function exists for: a \w filter collapsed the whole
+    // title to a single underscore, so every Chinese session exported as _.json.
+    expect(filenameStem('会话导出测试')).toBe('会话导出测试')
+  })
+
+  it('keeps other non-Latin scripts and accents', () => {
+    expect(filenameStem('Отладка платежей')).toBe('Отладка платежей')
+    expect(filenameStem('Rückläufer prüfen')).toBe('Rückläufer prüfen')
+    expect(filenameStem('日本語のタイトル')).toBe('日本語のタイトル')
+  })
+
+  it('drops characters a filename cannot hold', () => {
+    expect(filenameStem('a/b\\c:d*e?f"g<h>i|j')).toBe('abcdefghij')
+  })
+
+  it('strips control characters', () => {
+    const withControl = 'order' + String.fromCharCode(1) + 'sync'
+    expect(filenameStem(withControl)).toBe('ordersync')
+  })
+
+  it('collapses whitespace runs and newlines', () => {
+    expect(filenameStem('order   sync\nlater')).toBe('order sync later')
+  })
+
+  it('drops leading dots so the file is not hidden', () => {
+    expect(filenameStem('...hidden')).toBe('hidden')
+  })
+
+  it('drops trailing dots and spaces, which Windows discards silently', () => {
+    expect(filenameStem('report...')).toBe('report')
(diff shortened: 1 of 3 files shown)


Leihb:
Added the CJK filename fix here (c7eed2b2), since it is the same seam this PR already touches — both format functions compute the filename right where `deliverExport` was introduced.

**The bug:** `title.replace(/[^\w.-]+/g, '_')` keeps only ASCII word characters, so a title with none collapsed to a single underscore. Every Chinese session exported as `_.md` / `_.json`. Titles are auto-generated from the conversation, so this hit every non-English user on every export. I flagged it as out of scope when opening this PR; Roy hit it in the desktop build, so it is in scope now.

**The fix:** a `filenameStem` module that removes only what a filename genuinely cannot hold — Windows-reserved punctuation (a superset of POSIX's `/` and NUL) and control characters — and keeps everything else. Plus the things that bite once an OS save dialog is involved: a leading dot hides the file on POSIX, Windows silently discards trailing dots and spaces, and the length cap counts code points rather than UTF-16 units so a truncation cannot leave an unpaired surrogate.

**One ordering detail worth a reviewer's eye:** whitespace collapses *before* control characters are stripped. A newline is itself a control character, so doing it the other way round welds the words either side of it together — `sync\nlater` became `synclater`. A test caught that in the first draft.

**Verified** through the real UI in Chrome via CDP, both delivery paths:

```
desktop (native save dialog):  导出测试会话.md    导出测试会话.json
browser (blob download):       导出测试会话.md    导出测试会话.json
```

11 new unit tests cover CJK, Cyrillic, accents, Japanese, the illegal-character set, control characters, leading/trailing dots and spaces, the fallback when nothing survives, the code-point cap, and the surrogate-pair case. Full suite: 155 passed across 13 files. `svelte-check` 0 errors.

PR #2067, opened 2026-08-07
feat(tools): convert fetched HTML to readable Markdown in web_fetch

Leihb:
## Summary

Implements `dev-docs/web-fetch-cleaner.md` (#2064). Dropping the Jina proxy (#2065) left `web_fetch` returning raw HTML; this brings back "clean readable text" locally, with **no new module**.

Pipeline: `read → decode to UTF-8 → extract main content → convert to Markdown → existing size limits`. New `clean` param (default `true`) turns conversion off for callers that want the markup itself.

## What this fixes, measured

| Page | Raw | Cleaned |
|---|---|---|
| go.dev/blog/go1.22 | 36 KB | 3.4 KB |
| MDN `Referer` | 174 KB | 2.3 KB |
| Wikipedia *Go* | 720 KB | 101 KB (spills, and its outline works again) |

The spill outline is the quiet one: `markdownOutline` only recognises ATX headings, so post-Jina every large page spilled to a file with an empty outline and a preview of 40 lines of `<head>`. Cleaned output restores it.

## Extraction

Trimmed readability — prose elements score their parent and grandparent, the winner is penalised by link density and nudged by class/id, then **promoted upward while each parent holds ≥50% as much prose**.

That last step is load-bearing. Without it a reference-heavy article loses to its own least-linked section: Wikipedia's Go article cleaned to 13 KB starting at "Types", with the lead and History silently gone. The 50% threshold is where parent scores fall off a cliff across the sample pages — 0.35 and 0.25 produce byte-identical output on all six, so it's a structural boundary rather than a tuned constant.

## Decisions worth knowing

- **Charset decoding runs regardless of `clean`.** `html.Parse` assumes UTF-8, so a GBK page decodes to garbage *silently* — nothing errors, so no fallback fires. "Raw" means unextracted, not undecoded.
- **Links resolve against `resp.Request.URL`** (the final hop, so a shortener doesn't break every link), with `<base href>` taking priority. The spill filename and UI card keep the caller's original URL.
- **Output always leads with the page title**, which lives in the `<head>` that extraction never sees.
- **Tables** convert to GFM only when regular; `colspan`/`rowspan`/nested degrade to one line per row. Content survives, alignment doesn't.
- **The whole-page fallback deliberately keeps nav and links** — a page that fails extraction is usually a directory whose links *are* the content.
- **A non-boolean `clean`** is treated as absent, not rejected.

## Known limitation

`<table>`-layout pages (Hacker News' front page) collapse into very few lines. Every item, link and score survives and the model can use it, but it reads badly. Fixing it needs layout-vs-data table detection; documented in the design doc rather than guessed at.

## Also here

`stripHTML` (web_search) moves to `html.UnescapeString` — the hand-rolled replacer knew six entities and passed the rest through verbatim. Tags still go through the regex.

## Testing

`go test -race ./...` green. `gofmt -l .` clean, `go vet ./...` clean.

24 new test cases covering extraction, the fallback, GBK via header and via `<meta>` (and under `clean=false`), relative/`<base>`/post-redirect URL resolution, GFM and degraded tables, fenced code, the >64 KB spill-with-outline path, and non-boolean `clean`.

Beyond the fixtures, the extraction thresholds were calibrated against six real downloaded pages (go.dev, MDN, Wikipedia, pkg.go.dev, HN, a Zhihu anti-bot stub) — that sweep is what produced the 50% figure and confirmed the 200-char floor.


Leihb:
## Post-review fixes (isolated reviewer pass)

An isolated review found **no Critical**, one **Important**, and a handful of minors. Fixed in `ebb07104`:

- **`<base href="javascript:...">` scheme smuggling (Important, verified)** — `resolve()` checked the scheme of the raw href, but `ResolveReference` runs after that check, so a poisoned `<base>` turned every relative link into `javascript:///...` in the Markdown output. The whitelist now runs on the **resolved** result (absolute http/https only), and base adoption itself refuses a non-http(s) base so relative links resolve against the response URL instead of being dropped.
- `)` in emitted URLs is `%29`-escaped (Wikipedia article names hit this innocently — an unescaped paren closes the Markdown link early).
- Inline `<code>` containing backticks gets a lengthened delimiter, same rule as fences.
- Fence language hints from `class=` are restricted to token-safe characters.
- Comment documenting `decodeToUTF8`'s re-cap of the decoded size (legacy encodings grow ~1.5× as UTF-8).

New tests: malicious `<base>` end-to-end + at the resolver, all four `isHidden` branches (+ `aria-hidden=false` staying visible), paren escaping, backtick spans, fence-token rejection.

**Noted, not changed** (reviewer agreed these are acceptable):
- Flat pages (`<p>` directly under `<body>`) never score a candidate and always take the whole-page fallback — content survives, extraction quality gap only; real pages virtually always have a wrapper.
- Markdown injection via text nodes (a page can fake headings or mimic tool-output text) is inherent to fetching untrusted pages and predates this change — the raw-HTML path had the same exposure. Structure-breaking vectors (table pipes, fence collisions, link parens) are all escaped.
- Reviewer also measured the worst-case DOM walk: 400-level nesting × 4.5 MB ≈ 1.0 s / ~300 MB peak; parser depth cap (~512) bounds recursion. Not a DoS surface.

PR #2083, opened 2026-08-09
feat: add xAI (Grok) as a first-class vendor preset

### LINKED ISSUE #1872: Add xAI (Grok) as a first-class vendor preset
> 中文读者：**适合新人的 Go 小任务。** xAI 现在已经能用（走 Custom + openai 协议手填
> base URL），缺的是"一等公民"预设——选完 vendor 就有模型下拉、不用记 base URL。
> 改动集中在一个 Registry 数组。想认领的在下面留言。

## Problem

xAI (Grok) is reachable today, but only the hard way: pick the **Custom** vendor, set the protocol to `openai` by hand, and type the base URL and model id yourself.

Every other major provider is a first-class entry in `internal/app/provider.go`'s `Registry` — pick it in `octo config` or the settings panel and you get the right base URL plus a model dropdown. xAI is not there, even though the project already knows about Grok in two places:

- `internal/app/provider.go` lists `x-ai/grok-4.5` / `x-ai/grok-4.3` — but only inside the **OpenRouter** catalogue
- `internal/agent/compaction.go` already has Grok context-window handling

So the model is half-known to the codebase; it just has no vendor of its own.

## Goal

Add xAI as a first-class vendor:

1. A `Vendor` entry in `Registry` with `Protocol: "openai"`, the correct default base URL, and a model catalogue with accurate `Vision` flags per model.
2. Verify Grok's context windows in `internal/agent/compaction.go` are right for the models you add (the entries there predate this and may need updating).
3. Any i18n label keys needed, in **both EN and ZH** (`web/src/lib/i18n.ts`).

Look at the existing DeepSeek / Minimax / GLM entries as templates — DeepSeek is the simplest, Minimax shows the `BaseURLVariants` pattern if xAI needs regional endpoints (it probably does not).

## Please verify against xAI's current docs

Do **not** copy the base URL or model list from this issue or from memory — check xAI's official API docs when you write the PR, and say in the PR description what you verified and when. Model ids and vision support change.

Two specific things to confirm:

- Whether xAI's endpoint is fully OpenAI-compatible for the parts octo uses: streaming SSE, tool calls, and `stream_options.include_usage` (octo sends it — see the note in `CLAUDE.md` about backends that omit or reject the usage chunk).
- Which Grok models accept image input, so `Vision` is accurate rather than optimistic.

## Acceptance criteria

- [ ] xAI appears as a selectable vendor in `octo config` and in the Web settings panel
- [ ] Default base URL is correct; the user does not have to type it
- [ ] Model dropdown lists current Grok models with accurate `Vision` flags
- [ ] Context-window handling in `compaction.go` matches the models added
- [ ] i18n keys added for EN and ZH
- [ ] A real-key smoke test result described in the PR (run by hand — **no live network in `go test`**, see `CONTRIBUTING.md`)

## Notes

- `make test && make vet && make fmt-check` before pushing.
- One concept per PR: this is the vendor preset only. If you find a genuine wire-format incompatibility that needs adapter changes, raise it here first rather than bundling it in.

Leihb:
## What

Add xAI (Grok) as a first-class vendor preset in `internal/app/provider.go`'s Registry. Picking **xAI** in `octo config` or the Web settings panel now fills in the base URL, the API-key env var, and a Grok model dropdown — no more Custom vendor + hand-typed OpenAI endpoint.

Changes:
- **`internal/app/provider.go`** — new `xai` vendor entry: `Protocol: "openai"`, `DefaultBaseURL: https://api.x.ai`, 5-model catalogue, `LiteModel: grok-4.20-non-reasoning`, `APIKeyEnvVar: XAI_API_KEY`, key page link to `https://console.x.ai/`.
- **`internal/agent/compaction.go`** — Grok context windows corrected/added to match the xAI catalogue (see below).
- **`internal/app/provider_test.go`**, **`internal/agent/compaction_test.go`** — tests pinning the new vendor fields, base URL, env var, vision flags, and context windows.

CLI (`octo config`), completion, and the Web settings panel all derive from `app.Registry`, so no other code changes were needed. No new i18n keys were required — vendor labels flow straight from the registry `DisplayName` (the Web provider dropdown renders `p.name` directly), matching how every other vendor label works.

## Docs verified (xAI official)

Checked against **https://docs.x.ai** on **2026-08-09** (model pages show "Last updated: July 9, 2026"; model catalogue data taken from the site's published model registry):

- **Base URL**: `https://api.x.ai` — official OpenAI-SDK example uses `base_url="https://api.x.ai/v1"`; octo stores the host only and appends the protocol path, so the entry uses `https://api.x.ai`.
- **Vision flags**: the xAI model catalogue lists `inputModalities: ["TEXT", "IMAGE"]` for **all** current chat models (docs also state image input: jpg/jpeg, png, up to 20 MiB, unlimited count). Hence `Vision: true` for every catalogue entry. `grok-4.20-multi-agent` is **excluded** deliberately: it targets xAI's multi-agent orchestration API rather than a plain OpenAI-compatible chat client.
- **Context windows** (catalogue `maxPromptLength`):
  - `grok-4.5` → 500,000 (unchanged; was already correct)
  - `grok-4.20` → 1,000,000 (was mis-classified under `grok-4` = 256k)
  - `grok-4.3` → 1,000,000 (was mis-classified under `grok-4` = 256k)
  - `grok-build-0.1` (alias `grok-code-fast-1`, the code model) → 256,000 (was falling through to `grok` = 64k)
- **OpenAI compatibility for the parts octo uses**:
  - **Streaming SSE**: supported — official SDK/curl streaming examples use `stream=true` with `text/event-stream`.
  - **Tool calls**: supported — function calling is enabled on all listed models; client-side function tools return `tool_calls` for the caller to execute (docs "Tool Use").
  - **`stream_options.include_usage`**: **supported** — xAI's docs explicitly show `stream_options: {"include_usage": true}` and state usage is included in the final chunk (with empty `choices`), which matches how octo's OpenAI adapter consumes it. No wire-format or adapter changes needed.

## Smoke test

Not run — no xAI API key available in this environment. **Smoke test pending manual execution** with a real key: create an endpoint with vendor `xai`, send a chat turn, and confirm the usage/token accounting shows up. The code paths involved (registry lookup → openai client build → streaming) are covered by unit tests; only the live-API round trip is unverified.

## Notes

- No unrelated changes bundled. If a wire-format incompatibility with xAI surfaces during smoke testing, it will be tracked separately.
- `make test` (race), `make vet`, `make fmt-check` all green locally.

Closes #1872

### DIFF
--- a/internal/agent/compaction.go
+++ b/internal/agent/compaction.go
@@ -192,9 +192,17 @@ func contextWindow(model string) int {
 		return 64_000
 
 	// ── XAI Grok ──
+	// Context windows per xAI's model catalogue (docs.x.ai, 2026-07-09):
+	// grok-4.5 = 500k, grok-4.20 / grok-4.3 = 1M, grok-build-0.1 = 256k.
 	case strings.Contains(m, "grok-4.5") || strings.Contains(m, "grok4.5"):
 		return 500_000
+	case strings.Contains(m, "grok-4.20") || strings.Contains(m, "grok4.20"):
+		return 1_000_000
+	case strings.Contains(m, "grok-4.3") || strings.Contains(m, "grok4.3"):
+		return 1_000_000
 	case strings.Contains(m, "grok-4") || strings.Contains(m, "grok4"):
+		return 256_000
+	case strings.Contains(m, "grok-build") || strings.Contains(m, "grokbuild"):
 		return 256_000
 	case strings.Contains(m, "grok"):
 		return 64_000
--- a/internal/agent/compaction_test.go
+++ b/internal/agent/compaction_test.go
@@ -126,6 +126,23 @@ func TestContextWindow(t *testing.T) {
 	if got := contextWindow("x-ai/grok-4.5"); got != 500_000 {
 		t.Errorf("grok-4.5 window = %d, want 500000", got)
 	}
+	if got := contextWindow("grok-4.20"); got != 1_000_000 {
+		t.Errorf("grok-4.20 window = %d, want 1000000", got)
+	}
+	if got := contextWindow("grok-4.20-non-reasoning"); got != 1_000_000 {
+		t.Errorf("grok-4.20-non-reasoning window = %d, want 1000000", got)
+	}
+	if got := contextWindow("grok-4.3"); got != 1_000_000 {
+		t.Errorf("grok-4.3 window = %d, want 1000000", got)
+	}
+	// The OpenRouter-prefixed id resolves by substring and shares the model's
+	// real window — regression pin for the grok-4 (256k) → grok-4.3 (1M) change.
+	if got := contextWindow("x-ai/grok-4.3"); got != 1_000_000 {
+		t.Errorf("x-ai/grok-4.3 window = %d, want 1000000", got)
+	}
+	if got := contextWindow("grok-build-0.1"); got != 256_000 {
+		t.Errorf("grok-build-0.1 window = %d, want 256000", got)
+	}
 	if got := contextWindow("xiaomi/mimo-v2.5"); got != 1_000_000 {
 		t.Errorf("mimo-v2.5 window = %d, want 1000000", got)
 	}
--- a/internal/app/provider.go
+++ b/internal/app/provider.go
@@ -140,6 +140,28 @@ var Registry = []Vendor{
 		WebsiteURL:   "https://openrouter.ai/keys",
 	},
 	{
+		ID:             "xai",
+		DisplayName:    "xAI (Grok)",
+		Protocol:       "openai",
+		API:            "openai-completions",
+		DefaultBaseURL: "https://api.x.ai",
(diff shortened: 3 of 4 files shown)


Leihb:
**Review follow-up** (addressing review feedback):

- Added regression pins in `compaction_test.go` for `grok-4.20-non-reasoning` (the LiteModel) and `x-ai/grok-4.3` (OpenRouter-prefixed id whose window changed 256k → 1M, now explicit).
- Added a `LiteModel` assertion to `TestVendor_XAI` (value + membership in `Models`).
- Context-window figures (grok-4.5=500k, grok-4.20/grok-4.3=1M, grok-build-0.1=256k) were extracted from xAI's official model catalogue JSON embedded in docs.x.ai (`__XAI_PUBLIC_MODELS__`, `maxPromptLength` field; model pages dated July 9, 2026), not from memory — noted in the PR description above.

PR #2110, opened 2026-08-13
fix(tasks): close the plan out at turn end instead of trusting the model

### LINKED ISSUE #2102: agent 宣称任务完成但未调用 task_update，计划面板残留 in_progress 任务
## 问题描述

多步任务场景下，agent 完成了所有实际工作，并在最终回复中宣称「任务 2 完成，全部完成」，但忘记调用 `task_update` 把任务标记为 `completed`。结果是 Web UI 的计划面板里仍挂着一条 in_progress 的任务（例如显示 `计划 1/2 done`），与 agent 的文字汇报不一致，用户看到的状态是错的。

实际案例：一个两步任务（push demand + 更新提测评论），评论已成功添加（comment 114497），agent 回复「评论已添加。任务 2 完成。全部完成。」，但计划面板中「更新提测评论」一条仍是进行中状态。

## 根因分析

- 系统 prompt 已要求 agent「随手更新任务状态」（update status as you go / do not batch up），但这依赖模型的指令遵循，长上下文或多工具调用后容易在最后一步遗漏。
- 用户侧无法通过 hooks 兜底：`Stop` 事件是 side-effect only，stdout 会被丢弃，hook 无法在 turn 收尾时把提醒注入回模型上下文（见 hooks 文档：https://octo-agent.dev/docs/guides/hooks/）。

## 建议方案

在 agent loop 的收尾阶段（turn 结束前）加一个自动守卫：

1. turn 结束时检查 session 的 task list；
2. 如果仍存在 `in_progress` 状态的任务，注入一条 system-reminder（例如「以下任务仍为 in_progress，请确认并调用 task_update 更新状态」），让模型补调 `task_update` 后再真正结束 turn；
3. 可配置兜底策略：注入提醒一次后仍未更新的，可选择自动标记或保持原样并在 UI 上提示。

这样不依赖模型自觉，从机制上保证「文字汇报状态」与「任务面板状态」一致。

## 替代/缓解方案（已在用户侧验证可行）

在跨 session memory 的「必须遵守」区写一条强规则（每轮以 system-reminder 注入）：收尾前必须检查 task_list，宣称完成的任务必须先 `task_update` 再发最终回复。能缓解但不治本，仍建议实现上述守卫。

Leihb:
Fixes #2102.

## The problem

A multi-step turn routinely ends with the model reporting "all done" in prose while the last task is still marked `in_progress`. The system prompt already tells it to update status as it goes, but after a long tool-heavy turn the closing `task_update` is exactly the thing that gets dropped — so the plan panel shows `1/2 done` against a reply that says everything is finished.

Users can't backstop this themselves: the `Stop` hook is side-effect only, its stdout is discarded, so nothing a hook prints can reach the model.

## The fix

A turn-end seam in the agent loop. `Agent.TurnEndReminder` is consulted at the moment a turn would return — the model answered in prose, the inbox is empty, `EventTurnDone` is about to fire. A non-empty return is appended as a user message and the loop runs one more round, so the model can fix the record *inside* the same turn.

Three guards:

- **Fires at most once per turn.** A model that ignores the reminder must not be able to hold the turn open.
- **Stands down on the last allowed iteration.** Spending the final round on bookkeeping would turn a clean finish into a `max-turns` stop.
- **The agent layer stays ignorant of what is checked.** The ctx handed to the callback is the running turn's, so a ctx-scoped store resolves the same way the tools do.

`tools.PendingTaskReminder` is the first user of that seam: it names the tasks still marked `in_progress` and asks for the closing `task_update` (or `deleted`, if the task no longer applies).

**It fires only when nothing is pending.** An `in_progress` task with work still queued behind it is a plan the model is legitimately in the middle of — it commonly stops there to ask the user something — and nagging about that would cost an extra round-trip on every such turn. With the queue empty, the model believes it is on the final step, which is exactly the case this guard exists for.

The note is a `<system-reminder>`, so no UI renders it (a user message that strips to empty is skipped by web replay and the TUI alike).

## Wiring

All three transports, since each builds its tool environment differently:

| Transport | Site |
|---|---|
| CLI | `app.WireTools` (only when tasks are enabled; cleared on cleanup) |
| Web / SDK | `app.NewSessionToolEnv` (covers all three `prepareToolTurn` callers) |
| IM | `runChannelTurns` — this path assembles its tool env by hand |

Sub-agents are deliberately left out: they share the parent's checklist and must not report on it.

## Not done

The issue floats an "auto-mark it completed if the reminder is ignored" fallback. Marking a task done that the model never confirmed replaces a stale panel with a lying one, which is worse than the bug. No config knob either — the guard only costs a round-trip when there is an `in_progress` task and nothing pending, which is too narrow to be worth a setting.

## Tests

- `internal/agent/turn_end_reminder_test.go` — injects and continues / empty return ends the turn / once per turn and re-arms next turn / stands down at the turn limit.
- `internal/tools/task_guard_test.go` — 7-case status table (including the reported 1-done-1-in_progress shape and the mid-plan case that must stay quiet), only open tasks are named, ctx-store resolution.
- One wiring assertion per transport (`app/bootstrap_test.go` extended, new `app/toolenv_test.go` and `server/channel_task_guard_test.go`).

`go test -race ./...`, `go vet ./...`, `gofmt -l .` all clean.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -334,6 +334,23 @@ type Agent struct {
 	// so runLoop can detect when the model is stuck repeating the same call(s).
 	// Guarded by the implicit serialisation of runLoop (single goroutine).
 	recentToolCalls [][]toolCallFingerprint
+
+	// TurnEndReminder, when set, is consulted at the moment a turn would end:
+	// the model answered in prose and nothing else is pending. A non-empty
+	// return is appended as a user message and the loop runs one more round, so
+	// the model can act on it before the turn really closes; "" ends the turn.
+	//
+	// It is the seam for harness-side end-of-turn bookkeeping the model tends
+	// to forget — today the task-checklist guard (tools.PendingTaskReminder),
+	// which catches a plan left with an in_progress task after the model has
+	// already reported the work done. The agent layer stays ignorant of what
+	// is being checked; the ctx is the running turn's, so a ctx-scoped store
+	// (server / IM) resolves the same way the tools do.
+	//
+	// Fired at most once per turn: a model that ignores the reminder must not
+	// be able to hold the turn open. Wired only on top-level agents — a
+	// sub-agent shares the parent's checklist and must not report on it.
+	TurnEndReminder func(ctx context.Context) string
 }
 
 // toolCallFingerprint is a lightweight hash of a single tool-use block.
@@ -982,6 +999,7 @@ func (a *Agent) runLoop(
 	streamStalls := 0      // transient mid-stream stalls re-issued for the current round
 	truncationResumes := 0 // layer-2 resume-and-chunk budget
 	escalateExhausted := false
+	reminded := false // TurnEndReminder already fired this turn
 	a.turnIterations = 0
 	for i := 0; limit == unlimitedTurns || i < limit; i++ {
 		a.turnIterations = i + 1
@@ -1234,6 +1252,21 @@ func (a *Agent) runLoop(
 		// once-only — it fires only when we actually return below.
 		if a.Inbox.HasPending() {
 			continue
+		}
+
+		// End-of-turn bookkeeping the model forgot (a task left in_progress
+		// after it reported the work done). Inject the reminder as a user
+		// message and run one more round so it can fix the record inside this
+		// turn — the once-per-turn latch keeps a model that ignores it from
+		// holding the turn open.
+		// (Skipped on the last allowed iteration: the extra round would run out
+		// of loop budget and turn a clean finish into a max-turns stop.)
+		if !reminded && a.TurnEndReminder != nil && (limit == unlimitedTurns || i+1 < limit) {
+			if note := a.TurnEndReminder(ctx); note != "" {
+				reminded = true
+				a.History.Append(NewUserMessage(note))
+				continue
+			}
 		}
 
 		if handler != nil {
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -339,6 +339,9 @@ type Agent struct {
 	// the model answered in prose and nothing else is pending. A non-empty
 	// return is appended as a user message and the loop runs one more round, so
 	// the model can act on it before the turn really closes; "" ends the turn.
+	// toolsUsed carries the tool names dispatched so far this turn, so a
+	// reminder can scope itself to turns that actually touched what it guards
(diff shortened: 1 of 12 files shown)


Leihb:
Isolated review pass — three findings fixed in 753df8d1.

**1. The extra round could erase the answer on the web UI.** `EventTurnDone` broadcasts `Reply.Content` as the settled content of the streaming block (`ws_handlers.go` → the `assistant_message` overwrite branch in `ChatView.svelte`). When the model answers the reminder in prose — which the reminder's own wording encourages — both texts stream into that one block, so shipping only the second half replaced the answer with the bookkeeping line. History kept both, so a page refresh brought it back: live and reload disagreed.

Fixed by carrying the pre-reminder answer and re-attaching it to the final reply, with the carry dropped when a tool call or a user-visible steer opens a new block in between. TUI and IM were never affected (they render their own streamed buffers, not `Reply.Content`). Side benefit: the Stop hook now reports the whole stretch instead of just the bookkeeping line.

**2. The guard widened the interrupt window.** It opened a fresh provider round-trip between the model's answer and the return, so a cancel landing in that gap replaced an already-delivered answer with the interrupt note. Now skipped when `ctx.Err() != nil`.

**3. Unbounded re-billing across turns.** An unfinished plan survives across turns by design, and the reminder explicitly lets the model answer "that task really isn't done" — so one task it refused to close taxed *every* later turn in the session, including turns about something else entirely. `TurnEndReminder` now receives the turn's tool names and the guard only fires on turns that actually ran a `task_*` tool. That also matches the reported bug exactly (the model *did* call `task_update` that turn, just not the last one), and it stops the reminder naming `task_update` on a turn where the active profile never advertised it.

Also folded in: `NewSessionToolEnv`'s cleanup now clears the guard so an SDK caller reusing one agent across sessions doesn't inherit it, and both copies of the "Contracts" doc block (`internal/app` + the public `pkg/octoagent` mirror) — which claimed the function writes no `Agent` fields — are updated.

Declined: auto-marking a task completed when the reminder is ignored (replaces a stale panel with a lying one), and a config knob (the trigger is now narrow enough not to need one).

New tests cover the carry (kept / dropped by a tool call / dropped by a visible steer / kept through an invisible background-completion steer), the cancelled-turn skip, the tool-scoping table, and that a sub-agent never inherits the guard. `go test -race ./...`, `go vet ./...`, `gofmt -l .` clean.

PR #2141, opened 2026-08-17
feat: sub-agent "lite" model keyword + endpoint model list in schema

Leihb:
## What

Two additions to sub-agent model selection:

**1. `"lite"` as a model-override value.** The sub_agent tool's `model` parameter, the workflow `agent(model:)` opt, and agent-profile frontmatter `model:` now accept the special value `lite` (case-insensitive). At spawn time it resolves to the parent's lite sender/model (the endpoint's `lite_model`); when no lite model is configured it falls back to the parent's model. Resolution lives in one place — `app.Spawner.Spawn` — which every spawn path (sub_agent tool, workflow agents) already converges on.

This does not change the settled "no silent downgrade" policy: built-in presets still always inherit the parent's model. `lite` is an explicit per-preset or per-call opt-in for mechanical subtasks.

**2. Endpoint model list in the tool schema.** Previously the `model` parameter description was just "Optional model override" — the model had no way to know which override names are valid, so the parameter was effectively unusable unless the user dictated a name. `DefaultToolsFor(model)` already carries the session model, so the sub_agent definition is now built per-session: the description lists the sibling models of the session model's endpoint (from the cached config, lite model marked `(lite)`). Only same-endpoint models are listed because the child reuses the parent's endpoint connection — a model served elsewhere would be rejected.

## Notes

- Frontmatter parsers already pass `lite` through untouched (only `inherit` is normalized to empty), so no parser change was needed.
- When the session model isn't found in the config (ad-hoc `--model` value), the description falls back to the base text without a list.

## Tests

- `TestAgentSpawner_LiteModelOverrideUsesLiteSender` / `_LiteModelFallsBackToParentWhenUnconfigured`
- `TestSubAgentModelParamDescFor_ListsSiblingModels` / `_UnknownModelOmitsList`, `TestAgentToolDefinition_DocumentsLiteKeyword`
- Full `make test` (race) green.

### DIFF
--- a/internal/app/spawner.go
+++ b/internal/app/spawner.go
@@ -67,10 +67,18 @@ func (s *Spawner) Spawn(ctx context.Context, req tools.SpawnRequest) (tools.Spaw
 	childTools := filterChildTools(s.toolsFn(ctx), req.Tools, req.DisallowedTools, req.ReadOnly)
 
 	// The child runs on the parent's sender + model unless the request
-	// overrides the model explicitly. No preset downgrades to the lite model:
-	// a sub-agent's output gates the parent's next step, so cost is trimmed
-	// via the lean system prompt (below), never via model quality.
+	// overrides the model explicitly. No preset downgrades to the lite model
+	// SILENTLY — a sub-agent's output gates the parent's next step, so cost
+	// is trimmed via the lean system prompt (below) by default — but an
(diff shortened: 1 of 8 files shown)


Leihb:
Also rode along: dropped the stale `'plan' for read-only planning` claim from the sub_agent tool description (631d236f) — the plan preset no longer exists, so a model following the description would hit an unknown-type error.

PR #2178, opened 2026-08-20
fix(web): artifacts panel stays the artifacts panel off a session

Leihb:
## Summary
- The sidebar panel toggle (title bar and in-chat header) fell back to showing the Light Apps list when no chat session was selected — e.g. opening it from the Light Apps page itself. Removed that special case: both toggles now always open the Artifacts panel, which already has its own empty state for "nothing to preview" and covers this exactly the same way it covers a session with no artifacts yet.
- `panelContent === 'lightapps'` itself is untouched — LightAppsView's "打开" action on a specific app still uses it.

## Test plan
- [x] `npx svelte-check --threshold error` — 0 errors
- [x] `npx vitest run` — 342 passing
- [x] `make build`
- [x] Verified visually: opened Light Apps page, clicked the 制品面板 toggle — panel now shows Artifacts' empty state instead of the Light Apps list

### DIFF
--- a/web/src/components/layout/Header.svelte
+++ b/web/src/components/layout/Header.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { onMount } from 'svelte'
-  import { view, cmdkOpen, sidebar, nativeShell, panelContent, activeSessionId, settingsModalOpen } from '../../lib/stores'
+  import { cmdkOpen, sidebar, nativeShell, panelContent, settingsModalOpen } from '../../lib/stores'
   import { t } from '../../lib/i18n'
   import { ws, wsState } from '../../lib/ws'
   import { notificationsEnabled, setNotificationsEnabled } from '../../lib/notifications'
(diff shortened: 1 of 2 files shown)


Leihb:
Closing: main already behaves this way, so there is nothing left to merge here.

- `Header.svelte`'s `togglePanel` is now `panelContent.set('session')` with no `'lightapps'` fallback — that function was rewritten wholesale by the "each column carries its own chrome / the toggle lives in the column it reveals" work (#2189, #2190), which dropped the fallback along the way.
- The `hdr-btn` this PR touched in `ChatView.svelte` no longer exists; the chat header moved to `chatHeaderSnippet`, and `panelContent.set(id ? 'session' : 'lightapps')` went with it.

Every remaining `panelContent.set(` on main is `'session'` or `null`, except `LightAppsView.svelte`'s per-app "打开" action — exactly the one this PR meant to keep. Since `Header` renders for every view, opening the panel from the Light Apps page already lands on the Artifacts empty state.

Resolving the conflict would mean discarding the diff against a rewritten baseline and leaving an empty (or regressive, for the imports) change set, so this closes unmerged rather than being force-pushed onto current main.

PR #2186, opened 2026-08-20
fix(tools): find a just-installed pwsh that PATH lookup misses

Leihb:
Follow-up to #2185, which told the model to stop writing files through
PowerShell cmdlets because Windows PowerShell 5.1 defaults them to ANSI and
UTF-16. This closes the other half: making sure we actually run PowerShell 7
when it is installed.

## The gap

`resolvePowerShell` (internal/tools/sandbox.go) only consulted PATH, and it is
a `sync.OnceValue` — so a miss is cached for the entire process lifetime.

The installer walks straight into that. At `ssPostInstall`
(packaging/windows/octo.iss) it runs, in order:

1. `EnsurePowerShell7` — `winget install --id Microsoft.PowerShell`
2. `LaunchApp` — `Exec` of `octo-desktop.exe`

The launched app is a child of the installer, so it inherits the environment
block the installer captured at startup, before winget added
`C:\Program Files\PowerShell\7\` to the machine PATH. The desktop app is the
serve process, so the entire session right after install runs on 5.1 — with
exactly the encoding defaults #2185 exists to warn about — even though pwsh is
sitting on disk. It takes a later launch from Explorer, which handles
`WM_SETTINGCHANGE`, before the install takes effect.

## Fix

When `LookPath("pwsh")` fails, probe the MSI's fixed layout:
`%ProgramFiles%\PowerShell\7\pwsh.exe`. `ProgramW6432` is checked too, so a
32-bit octo (where `ProgramFiles` is `Program Files (x86)`) still finds the
64-bit install.

Fixing it here rather than in the installer means every install channel
benefits — scoop, `winget install octo`, a bare exe download — not just
`octo-setup.exe`.

Store-distributed pwsh needs nothing: it sits under `WindowsApps` behind an
execution alias that `LookPath` already resolves.

## Test

`TestProbePwshMSI` covers the not-installed, unset-roots, 64-bit-root-wins, and
`pwsh.exe`-is-a-directory cases. Path-shaped only, so it runs on every
platform rather than being Windows-gated.

`gofmt -l` clean, `go vet` clean, `go test ./internal/tools/` green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,7 +44,7 @@ Five-layer stack with one-directional dependencies:
    Provider wire quirks are encapsulated here — the agent layer never branches on protocol.
 
 5. **Tools (`internal/tools/`)** — concrete `ToolExecutor` implementations.
-   - `terminal.go` — current canonical example. Tool name `terminal` rather than `bash` because the implementation shells out via the platform shell — `sh -c` on macOS/Linux, PowerShell (`pwsh`, else `powershell`) on Windows — not a hard `/bin/bash` dependency. The shell is selected in one place: `shellCommand` in `sandbox.go`. The model is told which shell it's on via the environment context (`cmd/octo/envcontext.go`).
+   - `terminal.go` — current canonical example. Tool name `terminal` rather than `bash` because the implementation shells out via the platform shell — `sh -c` on macOS/Linux, PowerShell (`pwsh`, else `powershell`) on Windows — not a hard `/bin/bash` dependency. The shell binary is selected in one place — `executil.PowerShell()` — shared by the terminal tool, hook scripts, and clipboard capture so a session never straddles two PowerShells; `shellCommand` in `sandbox.go` is the single place the command is wrapped. The model is told which shell it's on via the environment context (`cmd/octo/envcontext.go`).
    - `DefaultRegistry` dispatches by tool name. `DefaultTools()` returns the set sent to the LLM (tools are on by default; `--no-tools` disables them).
 
 ## Adding capability
--- a/cmd/octo/clipimg_windows.go
+++ b/cmd/octo/clipimg_windows.go
@@ -14,15 +14,11 @@ import (
 )
 
 // captureClipboardImage reads an image from the Windows clipboard via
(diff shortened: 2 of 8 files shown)


Leihb:
Isolated code review found three things worth fixing; all three are addressed in d35cf352.

**PowerShell was resolved in three places, not one.** `internal/hooks/hooks.go` and
`cmd/octo/clipimg_windows.go` each had their own `LookPath("pwsh")`, so the probe only
reached the terminal tool — on the run right after install the terminal tool would have
used pwsh 7 while a hook script still ran on 5.1, with different `&&` support and
different file-cmdlet encodings. The selection now lives in `executil.PowerShell()`,
which all three already import (`hooks` cannot import `tools` — `tools` imports
`hooks`). `CLAUDE.md`'s claim that the shell is "selected in one place" was already
untrue before this branch; it is true now and the wording was corrected.

**`ProgramW6432` was unreachable.** It only differs from `ProgramFiles` inside a 32-bit
process, and `.goreleaser.yaml` builds windows/amd64 and windows/arm64 only — we ship no
32-bit Windows binary. Replaced with `ProgramFiles(x86)` (an x86 pwsh MSI beside a
64-bit octo is a case that can actually happen), matching the root list
`internal/browser/chrome.go` already probes.

**The probe wiring had no test.** The old test exercised the path helper only — deleting
the probe call from `resolvePowerShell` left every test green. The tests now drive
`powerShellIn` with PATH emptied: fallback to `"powershell"` when no root holds a pwsh,
the probed path when one does, and a pwsh on PATH still beating the MSI location. Removing
the probe call now fails the first of those (verified).

Also taken from the review: a non-absolute root is skipped, because `exec.Cmd` resolves a
relative `Path` against `cmd.Dir` — which `shellCommand` sets to the session working
directory — so a relative `ProgramFiles` could have run a `pwsh.exe` out of the user's
repo. And the hardcoded major-version directory is now called out in the comment as
something PowerShell 8 will need.

One review suggestion not taken: probing `%LOCALAPPDATA%` for a per-user MSI install. The
per-user layout is not something I can verify from here, and guessing a subdirectory that
turns out wrong buys nothing — worth adding once someone confirms it on a real per-user
install.

PR #2189, opened 2026-08-20
feat(web): drag the sidebar's edge to resize it, within bounds

Leihb:
The full sidebar was pinned at 256px. Long session titles and project-nested
rows had nowhere to go, and on a wide display the column could not earn more
room. It now has a grip on its divider.

- **Bounds:** 200px – 480px, clamped during the drag, default 256px unchanged.
  Below the floor a session title has no space left beside its timestamp; above
  the ceiling the sidebar starts eating the conversation it exists to navigate.
- **Persisted** in `localStorage` under `octo.sidebarWidth`. A stored width from
  a build with different bounds is clamped rather than discarded — the intent
  was "as wide as possible", so it becomes the new maximum.
- **Rail and hidden modes are untouched.** Both are fixed widths with nothing to
  drag, so the grip only renders in full mode.

Follows `ArtifactsPanel`'s existing handle: same mouse-capture shape, same
`col-resize` cursor and `user-select` lock, same persistence.

Two details worth noting:

- The collapse transition is suppressed for the duration of a drag. It applies a
  new width every `mousemove`, so a 0.32s ease would leave the edge lagging
  behind the cursor.
- The grip carries `--wails-draggable: no-drag`. It overlays the sidebar header,
  which is a window-drag region in the desktop shell, and `framelessDrag`
  decides by the mousedown target's computed value.

`.full` keeps a pixel width (now driven by the drag state) rather than switching
to `100%`, so content still gets clipped by the `<aside>` during the collapse
animation instead of reflowing through it — the same reason `.rail` pins its
own width.

## Verification

`svelte-check` clean (0 errors; the 2 warnings are pre-existing `line-clamp`
notes elsewhere). Driven over CDP against an isolated `octo serve` on a throwaway
HOME — never the live backend:

| step | result |
| --- | --- |
| default width | 256 |
| drag +400px | 480 (capped) |
| drag −600px | 200 (floored) |
| drag to 340, reload | 340 (persisted) |
| collapse toggle | 1px (border only) — fixed modes unaffected |

Screenshots at both bounds confirm the header row (toggle, logo, brand, search)
and the footer (Settings, version) still fit at 200px, and that content follows
the column out to 480px.

### DIFF
--- a/web/src/components/ArtifactsPanel.svelte
+++ b/web/src/components/ArtifactsPanel.svelte
@@ -3,6 +3,7 @@
   import { t } from '../lib/i18n'
   import { copyArtifact, downloadArtifact, imagePreviewError } from '../lib/artifact-actions'
   import { hydrateArtifact, lightAppSource, pathIsInside } from '../lib/artifacts'
+  import { CENTER_MIN } from '../lib/sidebarWidth'
   import * as api from '../lib/api'
 
   // ── Session artifacts (existing) ──────────────────────────────────────────
@@ -169,7 +170,6 @@
   // center column drops under its own minimum (mirrors the design mock).
   const PANEL_WIDTH_KEY = 'octo.panelWidth'
   const PANEL_MIN = 320
-  const CENTER_MIN = 460
   let panelEl = $state<HTMLElement | null>(null)
   let panelWidth = $state(readSavedWidth())
 
--- a/web/src/components/layout/Sidebar.svelte
+++ b/web/src/components/layout/Sidebar.svelte
@@ -6,6 +6,7 @@
   import { t, tr } from '../../lib/i18n'
   import { confirmDialog } from '../../lib/confirm'
   import { splitSections, swapWithinSection, parseSectionFold, type SectionFold } from '../../lib/sidebarSections'
+  import { SIDEBAR_MIN, SIDEBAR_MAX, CENTER_MIN, readSidebarWidth, saveSidebarWidth } from '../../lib/sidebarWidth'
   import { ago } from '../../lib/relTime'
   import { ws } from '../../lib/ws'
   import VersionBadge from './VersionBadge.svelte'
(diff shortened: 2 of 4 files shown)


Leihb:
Isolated code review turned up one crasher and one reachable layout bug. Both fixed in c9450340,
along with three of the smaller notes.

**Bare `localStorage` inside a `$state` initializer (crasher).** Storage *access* throws when a
browser blocks it, and `readSavedWidth()` ran during `Sidebar`'s init — so instead of falling back
to the default width, the always-mounted sidebar failed to mount and took the app tree with it.
`loadSections()`, 90 lines above in the same file, already had the try/catch and a comment
explaining why. Reading and writing are both wrapped now.

**A static 480px ceiling that ignored the room left over.** On a 900px window with the artifacts
panel at its default 420px, one drag squeezed the conversation column to nothing — `.main` is
`flex:1; min-width:0`, so it absorbs the squeeze silently rather than overflowing. The panel's own
grip already caps itself against `CENTER_MIN = 460` for exactly this reason; the sidebar now caps
from its side too, measuring the row on mousedown. `CENTER_MIN` moved into
`lib/sidebarWidth.ts` so the number isn't written down twice.

**A `mouseup` that never arrives** (swallowed by browser chrome or a cross-origin frame) left
`resizing` stuck true, the body stuck in `col-resize` with text unselectable, and the width still
tracking the mouse. A move event with no button held now ends the drag. A non-primary button no
longer starts one either.

**Comment correctness.** The note on the grip's `--wails-draggable: no-drag` implied the grip needed
it to escape the header's `drag`. It doesn't: the property inherits through the DOM and the grip is
the header's sibling, so it never picked up `drag`. The declaration is defensive and the comment now
says that instead of asserting a mechanism that isn't there.

Not taken, with reasons:

- **Pointer events for touch input.** Real gap on a touchscreen laptop, but this UI is mouse-driven
  and `ArtifactsPanel` is mouse-only too. Worth doing for both grips at once, not for one of them here.
- **Keyboard operability on `role="separator"`.** Same argument — it would make the two grips
  inconsistent. Width is a preference with a sensible default, so nothing is unreachable without it.
- **`onDestroy` cleanup of an in-flight drag.** `Sidebar` is mounted unconditionally by `App.svelte`;
  only its inner `{#if}` swaps on mode change, so the component cannot unmount mid-drag.
- **`framelessDrag` fighting over `document.body.style.cursor`.** Real, but it predates this change
  and belongs with a fix to the shared pattern rather than to one caller.

## Verification

`svelte-check` clean. Full suite 366 tests / 25 files green, including nine new ones in
`sidebarWidth.test.ts`. Deleting the try/catch fails two of them (checked).

Re-driven over CDP against an isolated `octo serve` on a throwaway HOME:

| window | step | result |
| --- | --- | --- |
| 1440 | default | 256 |
| 1440 | drag +500 | 480 — absolute ceiling |
| 900 | drag +500 | 440 sidebar / 460 main — center minimum wins over the ceiling |
| 900 | drag −600 | 200 — floor |
| 900 | reload | width persisted |
| 900 | right-button drag | width unchanged, cursor not left as `col-resize` |

One thing worth writing down for whoever tests web code next: this jsdom setup exposes **no**
`localStorage` at all (`window.localStorage` is `undefined`), so storage cases have to be stubbed
with `vi.stubGlobal` — which also means the absent-storage path is exercised for real, not simulated.

PR #2306, opened 2026-08-28
web: keep chat auto-scroll alive across pointer release and window switches

Leihb:
## Problem

During a running turn, switching to another window and back left the chat transcript parked mid-way; new content kept arriving below the fold and had to be scrolled to by hand. Reported as a long-standing web-UI bug.

## Root cause

`onPointerUp` — bound to `pointerup`, `pointercancel` **and** `window blur` — re-ran `onScroll()` to re-derive `stick` from the resting scroll position. That position is stale whenever content growth has outrun the ResizeObserver pin, and `onScroll`'s "not moving up" branch then sets `stick = nearBottom`, i.e. **false** as soon as the view is >80px above the bottom. Once latched off, nothing re-arms it (the RO's re-arm requires already being at the bottom).

Two ways to hit it, both part of a window switch:

1. **Pointer press while streaming** (e.g. clicking back into the chat to refocus the window). `interacting` blocks the RO for the whole press while the stream keeps appending; on release the re-derive sees the gap and unsticks. After a hidden period the gap is large and the activating click lands before the first RO frame, so this reproduced every time.
2. **Window blur** landing between a DOM append and the next RO frame (≤1 frame window, but streaming appends constantly) — same re-derive, same latch.

## Fix

- `onPointerUp` no longer re-derives `stick`. A gesture that moved the scroller already updated `stick` through its own `scroll` events; a press that didn't scroll is not a gesture. On release/blur it only clears `interacting` and, if still following, catches up on the growth the press held back.
- New `visibilitychange` handler: on wake, re-pin if following and resync `lastScrollTop`, so scroll events coalesced across the hidden period (scroll anchoring / clamping after a block collapsed) can't be read as the user scrolling away.

`onScroll` / `onWheel` (the #1187 / #1259 / #1979 logic) are untouched.

## Verification

Ported the effect verbatim into a standalone page and drove it in headless Chrome over CDP (two tabs in one window so the page really goes `document.hidden`):

| scenario | before | after |
|---|---|---|
| press on scroller → content grows 300px → release → more content | `stick=false`, stuck | pinned |
| blur in the same task as a 300px append → more content | `stick=false`, stuck | pinned |
| hidden tab: block above collapses, then +3000px, wake | pinned (Chrome fires RO before the coalesced scroll event) | pinned |
| hidden tab: +3000px, wake | pinned | pinned |
| regression: scrollbar drag **up** during streaming → release | disengaged | disengaged |

`svelte-check` 0 errors, `vitest` 634 passed.

### DIFF
--- a/web/src/views/ChatView.svelte
+++ b/web/src/views/ChatView.svelte
@@ -1399,17 +1399,46 @@ import QuestionModal from '../components/overlays/QuestionModal.svelte'
     }
     scroller.addEventListener('wheel', onWheel, { passive: true })
 
+    // Releasing the pointer used to re-run onScroll() to re-derive stick from
+    // the resting position. But the RO is blocked for the whole press while the
+    // stream keeps appending, so by release the view can sit well below the
+    // 80px band without the user having scrolled anywhere — and that re-derive
+    // latched stick off for the rest of the turn. Clicking back into the chat
+    // after a window switch (content grew while the tab was hidden, the
+    // activating click lands before the first RO frame) hit this every time.
+    // A gesture that moved the scroller already updated stick through its
+    // 'scroll' events; a press that didn't scroll is not a gesture. So on
+    // release only lift the block and catch up on whatever growth it held back.
     let interacting = false
     const onPointerDown = () => { interacting = true }
-    const onPointerUp = () => { interacting = false; onScroll() }
+    const onPointerUp = () => {
+      interacting = false
+      if (stick) scroller.scrollTop = scroller.scrollHeight
+    }
     scroller.addEventListener('pointerdown', onPointerDown)
     window.addEventListener('pointerup', onPointerUp)
     window.addEventListener('pointercancel', onPointerUp)
     // A drag released outside the window (e.g. the pointer is still down when
     // focus is stolen by alt-tabbing) never fires 'pointerup' on this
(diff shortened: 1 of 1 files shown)


Leihb:
Review follow-up (isolated reviewer): one Important finding confirmed and fixed in the second commit.

**Sequence that still latched stick off:** hidden tab → a block above collapses (scroll anchoring adjusts scrollTop) → content grows → the activating `pointerdown` reaches the page **before** `visibilitychange` → `onVisibility` skips the pin (`interacting` is true) and resyncs `lastScrollTop` → the coalesced scroll event fires with `top === lastScrollTop` → `onScroll`'s non-upward branch re-derives `stick = nearBottom` → false → `pointerup` sees `stick` false and does not pin.

Reproduced in the standalone harness (`scroll top=2000 last=2000 near=false` → stuck), fixed by early-returning from `onScroll` when `top === lastScrollTop` (a scroll event that did not move cannot be a gesture). All six harness scenarios pass after the change, including the scrollbar drag-up regression guard. `svelte-check` 0 errors, `vitest` 634 passed.

Two minor items from the review left as-is: (1) `pointercancel` on touch now writes `scrollTop` once at pan start — same class of race the RO already has, not new; (2) extracting the stick state machine into a testable `lib/` module is worth a follow-up PR given this block's history, but out of scope here.

PR #2396, opened 2026-09-12
computer: focus the target app before pixel input, address AX elements by id

Leihb:
## Summary
- Addresses 2 of the 5 gaps reported in #2395 from real computer-use runs (macOS Notes, Lightroom Classic): focus management (#1) and AX element addressing (#2). The other three (menu-bar noise, coordinate-system unification, drag primitive) are left for follow-up PRs.
- `left_click`/`right_click`/`double_click`/`type`/`key` gain an optional `app` parameter that brings the target app to the foreground first (`kAXFrontmostAttribute` on macOS; `SetForegroundWindow` + `AttachThreadInput` on Windows to work around the foreground-lock rule) — Octo's own window could otherwise hold focus and silently swallow pixel-channel input. The existing `ClickPid`/`TypeTextPid` per-process delivery was already known unreliable per its own doc comments (dropped by both SwiftUI and custom-drawn views), so real focus is the fix that actually works. `type`/`key` now also report the frontmost app afterward so a miss is visible instead of silent.
- `ax_tree` now prefixes every line with its `e<N>` traversal index; `ax_press`/`ax_set` accept that `id` instead of role+label matching — the only way to address an element with an empty or duplicate label (e.g. several sliders all labelled `"0"`, or an empty-label text area). Implemented as an index-based re-walk (`axWalk`/`axActByIndex`, shared with `AXTree`) rather than a cached native-pointer table, to avoid cross-call lifetime/invalidation complexity. Ids are stable as long as the app's tree and the `max_depth` used haven't changed since the dump.
- Elements with empty labels are no longer filtered out of `ax_tree` unless they're pure layout containers (`AXGroup`/`AXSplitGroup`/etc.) — those are exactly the elements that need an id to be reachable at all.

## Test plan
- [x] `go build ./...` — darwin (cgo), `GOOS=windows CGO_ENABLED=0` cross-build, `GOOS=linux CGO_ENABLED=0` stub build
- [x] `go vet ./...` on all three
- [x] `gofmt -l .` clean
- [x] `go test -race ./...` passes, including new unit tests for id parsing and negative-id validation
- [ ] Real GUI interaction (focus switching, `AttachThreadInput` on Windows) not yet exercised on a live desktop — build/unit-test verified only, needs a manual pass before merge

### DIFF
--- a/internal/computer/ax_darwin.go
+++ b/internal/computer/ax_darwin.go
@@ -122,24 +122,29 @@ func axTree(pid, maxDepth int) ([]AXElement, error) {
 // the same pid + maxDepth. This is the id-addressed counterpart to
 // axFindDo's role+label matching, for elements ax_tree shows with an empty
 // or duplicate label (axFindDo can't disambiguate those by contains/role).
-func axActByIndex(pid, maxDepth, target int, action func(C.AXUIElementRef) error) error {
+// Returns the matched element's digest (role/label/frame) alongside any
+// error so the caller can echo back what it actually hit — the one signal
+// that an index-addressed action landed on the right widget.
+func axActByIndex(pid, maxDepth, target int, action func(C.AXUIElementRef) error) (AXElement, error) {
 	var found bool
+	var matched AXElement
 	var actErr error
-	err := axWalk(pid, maxDepth, func(i int, el C.AXUIElementRef, _ AXElement) bool {
+	err := axWalk(pid, maxDepth, func(i int, el C.AXUIElementRef, e AXElement) bool {
 		if i == target {
 			found = true
+			matched = e
 			actErr = action(el)
 			return false
 		}
 		return true
 	})
 	if err != nil {
-		return err
+		return AXElement{}, err
 	}
 	if !found {
-		return fmt.Errorf("computer: no element with id e%d (re-dump ax_tree — the tree may have changed, or max_depth differs from the dump that produced this id)", target)
+		return AXElement{}, fmt.Errorf("computer: no element with id e%d (re-dump ax_tree — the tree may have changed, or max_depth differs from the dump that produced this id)", target)
 	}
-	return actErr
+	return matched, actErr
 }
 
 // axFind locates the first element matching role+contains (same rules as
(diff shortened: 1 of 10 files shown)


Leihb:
Ran an isolated code review (fresh subagent, no context from the implementation) against 7c90d4a4. Summary:

**Strengths**
- `axWalk` shared cleanly between `AXTree` (dump) and the id-addressed action path on both platforms — no duplicated traversal logic, and the id path is actually faster (short-circuits instead of building the full digest).
- Fully backward compatible — `app`/`id` are optional everywhere.
- CGO memory management in the two new C helpers is balanced (checked manually).
- `id` parsing tolerates all the input shapes a model might send without panicking, with dedicated validation tests.
- The re-walk-by-index tradeoff (vs. caching native pointers) is explicitly documented as a known limitation, not silently assumed.

**Issues found and fixed in 279fe93f:**
- **Important** — `windows.go` `activateApp` called `GetCurrentThreadId`/`AttachThreadInput`/`SetForegroundWindow` without `runtime.LockOSThread()`. Go's scheduler could migrate the goroutine mid-sequence, attaching the wrong thread's input queue and making the foreground-lock workaround silently ineffective. Fixed by following the same `LockOSThread` convention `uia_windows.go`'s `withCOM` already uses for COM apartment affinity.
- **Important (test gap)** — nothing protected the invariant that `ax_tree`'s printed `e<N>` id always equals the element's true index (not a "shown so far" counter), even though some elements are filtered from display. A future edit using the wrong counter would silently misaddress every id-based `ax_press`/`ax_set`. Fixed by extracting the render loop into a pure `renderAXTree` function with a dedicated regression test — no AX/UIA substrate needed to test it.
- **Minor** — id-addressed `ax_press`/`ax_set` didn't echo back which element (role/label) they actually hit. Added (`axDigestSuffix`) since the matched digest was already available for free from the walk.

No critical/blocking bugs. Still outstanding before merge: **no real GUI interaction has been exercised on a live desktop** — everything above is build/vet/unit-test verified only, per the earlier PR description.

PR #2412, opened 2026-09-14
feat(browser): record each gesture's effects and flag trailing no-op clicks for the distiller

Leihb:
Part 3 of #2406, implementing `dev-docs/browser-recording-effects-design.md` (#2407) with every proposed decision as-is. The design doc is rewritten in this PR from proposal to current state.

## What

- **`Effects` per gesture** (`internal/browser/effects.go`), in `events.json` only: `url_after` / `new_tab` / `download` derived from the navigate and download events the recorder already emits; `requests` counted by HTTP method from `Network.requestWillBeSent` on every instrumented session (top document, OOPIF frames, new tabs), filtered to XHR/Fetch/Document, attributed to the latest gesture within 1.5 s. Method only — no URLs. Events now carry `at_ms`.
- **`likely_noop`**: trailing click/enter gestures with no non-GET request, no URL change, no new tab, no download are marked; the first event that fails stops the backward walk. Tail only. Never deleted on its own.
- **Distiller**: `renderTrace` shows each event's `url=`, `effects:` and `[likely_noop]`; rule (2) tells the model to drop marked events unless the goal needs them.
- **Plan**: the marker rides onto compiled steps (`Step.likelyNoop`, unexported, never in YAML), is re-attached onto refined steps by selector (`backfillNoopHints`), and `SummarizeRecording` lists surviving flagged steps under "以下步骤没有改变页面状态…要保留吗？" for the user.
- Docs (en/zh guide) mention the evidence and the question block.

## Verified end to end with real models

`TestDistillE2E_RealModels` (`internal/app`, opt-in `OCTO_E2E_DISTILL=1`, uses the keys in `~/.octo/config.yml`, never in CI) records the #2406 demonstration in headless Chrome through the real recorder — 笔记管理 → 最新笔记 → 查看评论, then the strays 说点什么... → 2回复 → 取消 — and distils it with every configured model four ways. Two runs, three models (kimi k3 / deepseek-flash / qwen3.7-plus):

| model | goal+effects | goal only | effects only | neither |
|---|---|---|---|---|
| kimi-coding-plan::k3 | pass, pass | pass, pass | pass, FAIL | FAIL, FAIL |
| deepseek::deepseek-flash | pass, pass | pass, pass | FAIL, FAIL | FAIL, FAIL |
| bailian::qwen3.7-plus | pass, pass | FAIL*, pass | FAIL*, pass | pass, pass |

pass = the three intended clicks kept, the three strays dropped, refinement applied. FAIL without goal = the model kept 2回复 (expanding a reply reads like "checking comments" absent the goal). \* = first run judged by `label`, which qwen omits from its YAML; the judge now resolves by selector.

The design column (goal + effects) is 6/6. Note honestly: on this fixture "goal only" also cleaned the recording in the second run for all three models, so the marginal value of the effects over the goal is not demonstrated by the model output here; what the effects add regardless is the deterministic `likely_noop` marking and the "要保留吗？" question in the plan — the user-facing path that does not depend on the model at all. The recorder marks 查看评论 too (a GET-only tab at the tail), which the goal correctly overrides in every run.

## Tests

- `TestRecorderCapturesEffects` (Chrome): method counts per click, `url_after` on the navigation, exactly the two trailing clicks marked.
- `TestAttachEffectsAttribution`, `TestMarkLikelyNoopTailOnly`, `TestEffectsString`, `TestRenderTraceEffects`, `TestLikelyNoopReachesPlan`.
- `go test -race` on browser / tools / app green locally.

## Side finding (not fixed here)

Models routinely omit `label` from the refined YAML (qwen did on every step in one run). Labels drive replay's text anchoring and the plan's wording; `backfillAnchors` restores anchors but not labels. Worth a follow-up: backfill `Label`/`Hint` by selector the same way.


Leihb:
## Review fixes + live-model results across 5 runs

Pushed ea1b009b (review fixes: `Network.disable` on Stop, 1024-deep request subscription, 300 ms lead tolerance for the cross-goroutine timestamp race, tail-only invariant re-imposed after distillation, Enter-with-value stops the walk, key steps worded as key presses) and softened rule (2): marked events are *candidates* — drop when the goal clearly does not need them, keep when the goal names them or is vague.

Cumulative matrix, 3 models × 5 runs (pass = 3 intended clicks kept, 3 strays dropped, refinement applied):

| variant | pass / runs | notes |
|---|---|---|
| goal + effects | 14 / 15 | the 1 miss: deepseek reply truncated at 2048 max_tokens → #2415 |
| goal only | 12 / 15 | 1 provider timeout, 1 truncation, 1 kept 2回复 |
| effects only | 5 / 15 | kimi/deepseek keep 2回复 without a goal; **1 over-deletion** (deepseek dropped 查看评论) → why rule (2) was softened |
| neither | 4 / 15 | the #2406 report |
| vague goal + effects | 4 / 9 | qwen 3/3, kimi 1/3, deepseek 0/3 |
| vague goal only | 1 / 9 | |

Honest reading: the **goal** is the decisive signal for the model; effects are not shown to add to it on this fixture. What effects add for sure is model-independent: the deterministic `likely_noop` marking and the "要保留吗？" question in the plan, which also covers the no-model and distill-failed paths. Note the recorder also flags the wanted GET-only tail step 查看评论; only the goal keeps it, so the marker's wording now defers to the goal.

Two failure classes surfaced by the e2e are outside this PR: the distiller's 2048-token cap truncating/emptying reasoning-model replies (#2415), and models dropping `label` from refined steps (#2413).

PR #2444, opened 2026-09-16
feat(profile): add profile-scoped user data and active profile indicator

woodcoal:
## Summary

Add first-class named profiles for Octo persistent user data.

`--profile <name>` now selects an isolated user-data root:

- Default: `~/.octo`
- `--profile work`: `~/.octo-work`

The selected profile is applied consistently across CLI, web server, and desktop entry points. Named profiles are also visible in the shared Web/Desktop sidebar.

### Global CLI profile support

- Add global `--profile <name>` and `--profile=<name>` parsing in any command position.
- Validate profile names and reject traversal/path-like values.
- Strip the global flag before subcommand-specific parsers run.
- Preserve `OCTO_PROFILE` through daemon, supervisor, sandbox, and child-process execution.
- Add shell-completion support for `--profile`.
- Update CLI help text.

### Profile-scoped data root

- Introduce `internal/datahome` as the single source of truth for user data paths.
- Keep the default location unchanged at `~/.octo`.
- Route named profiles to `~/.octo-<profile>`.
- Migrate user-scoped persistent paths to `datahome.Dir()` / `datahome.Path()`, including:
  - config, sessions, agents, prompts, skills, workflows, memory, hooks, permissions, audit logs, trash;
  - MCP configuration and OAuth tokens;
  - IM channel configuration and credentials;
  - server PID/logs, uploads, tasks, session groups, Light Apps, browser recordings, and temporary tool data.
- Preserve project-local `<workspace>/.octo` behavior.
- Do not change shell-level paths such as `~/.local/bin/octo`.

### Web and desktop profile visibility

- Extend `GET /api/version` to return `profile` only for named profiles.
- Render a compact `work`-style badge beside the version in the shared sidebar footer.
- Omit the badge for the default profile.
- Add title and ARIA label to the badge.
- The shared Web UI means the indicator appears in both browser and desktop-shell views.

### Desktop profile support

- Add `octo-desktop --profile <name>` support.
- Parse the profile before desktop crash logging, `serve.env`, settings, or server startup.
- Make desktop settings and bundled `uv` use the selected profile data root.
- Keep desktop crash logs, server logs, PID files, config, and `serve.env` aligned through the existing profile-aware services.
- Return exit code `2` for invalid desktop profile input.

### Serve output

`octo serve` now prints the selected named profile and resolved data directory after startup, for example:

```text
octo server listening on http://127.0.0.1:8088 profile: work
data directory: /home/user/.octo-work
For the default profile, only the data directory is printed.
```

## Verification

```text
go test ./cmd/octo -run "TestRun_(Profile|Version)|TestPrintUsage"
go test ./internal/server -run "TestHandleVersion_(NoUpdateCheck|ProfileContract|CacheHeaders)
```


Leihb:
Nice work. This is a feature we want, and the migration is more careful than most PRs this size: `internal/datahome` is the right shape, the default profile is unchanged, and `OCTO_PROFILE` really does propagate through the daemon, supervisor, sandbox and terminal children, since every spawn site builds on `os.Environ()`.

Two things block merge.

**1. `~/.octo/bin` shouldn't be profile-scoped.**

It isn't user data. The installers stage machine-level tooling there at a fixed path (`packaging/macos/scripts/postinstall:23`, `packaging/windows/octo.iss:132`), but this PR moves all the readers to the profile root: `bundledBinDir()` in `internal/tools/sandbox.go:181`, `octoBinDir()` in `internal/tools/rgembed/rgembed.go:120`, the sandbox read root at `internal/sandbox/sandbox.go:68`, `bundledUvTarget()` in `cmd/octo-desktop/uv.go:44`.

So under `octo --profile work`, `DetectToolchain()` probes `~/.octo-work/bin`, finds no uv, and reports the Python toolchain missing. `withBundledBinPath()` goes no-op, so children lose it from `PATH` too. Every Python skill quietly stops working, and a CLI-only install has nothing to re-seed it. `rgembed` also re-extracts its own ripgrep per profile.

Keep `bin/` at the unscoped root.

**2. The new `--profile` completion doesn't work.**

`run()` calls `ConfigureFromArgs(args)` before the `switch`, but for `__complete` those args are the user's command line, not octo's own:

```
$ octo __complete octo --profile ""
config
doctor
init
...
```

`octo --profile <TAB>` offers subcommand names as profile names, and the `case "--profile":` you added at `completion.go:132` is unreachable. A typo escapes too:

```
$ octo __complete octo "--profile=../x" ""
octo: invalid --profile: profile must contain only letters, digits, ...
```

which under bash lands on the user's prompt. The new test calls `completionCandidates()` directly, so it stays green either way.

Skip `ConfigureFromArgs` for the internal passthrough commands, or stop stripping at the first non-flag token.

**Smaller things**

- `VersionBadge.svelte:238` hard-codes `Profile: ${profile}` in both `title` and `aria-label`. `web/src/lib/i18n.ts` has en and zh maps, and the file uses `$t()` a few lines up.
- `datahome.go:14` says `ProfileEnv` is set only by the flag, but `Dir()` reads the env on every call and `TestConfigureFromArgsPreservesInheritedProfile` guards that. Pick one.
- Docs still say `~/.octo` is the only location: `concepts/configuration-layers.md`, `concepts/sessions-and-history.md`, `guides/self-host.md`.
- `ConfigDir` in `internal/channel/config.go:15` has no references left.
- In `internal/mcp/oauth.go`, `"time"` ended up in the third-party import group.
- `trash.Dir()` returning `""` makes `ProjectDir()` a bare hash directory in the CWD. Only when `UserHomeDir()` fails, but it used to be `./.octo/trash/<hash>`.
- `--profile=` with no value silently means the default profile.
- `cmd/octo-desktop/main.go:134` drops the filtered args, so `os.Args` keeps the stray `--profile work`.

Fix 1 and 2 and I'll merge it.

Leihb:
@woodcoal 加我微信，我拉你进群交流吧

woodcoal:
已经基于最新的分支进行了调整

Leihb:
Both blockers are properly fixed. `datahome.BinDir()` is the right seam for the installer-managed binaries, and I confirmed the completion behaviour on a build of your branch:

```
$ octo __complete octo --profile ""        → no candidates, exit 0
$ octo __complete octo "--profile=../x" "" → chat flags, no error leak
$ octo version --profile=                  → "--profile requires a name", exit 2
```

I pushed df5bc93 with three follow-ups rather than send you another round.

The one that mattered: `prepareDesktopArgs` rejected unrecognised arguments and exited 2, but it runs ahead of `setupCrashLog`, and the comment right below it says why that's a problem — a GUI launch has no stderr (Windows is `-H windowsgui`, macOS starts from Finder). Anything that ever passed an argument would have left the app silently refusing to start with nothing written down. Unknown arguments are ignored again; an invalid `--profile` still exits, since that can only come from a terminal.

The other two: `__complete` now adopts the profile named on the line being completed, so `octo --profile work -c <TAB>` offers the work profile's sessions instead of the default one's (best-effort, errors ignored — the words still reach the routing unshifted). And the docs tables name `~/.octo` again, with the profile variant stated once below them, since almost every reader is on the default.

Merging. Thanks for the quick turnaround on the first round.

PR #2445, opened 2026-09-17
feat(privacy): let the update check be turned off

Leihb:
## Why

README and the landing page claim octo makes no outbound request of its own. It does: the latest-release lookup against GitHub.

- `internal/server/version_upgrade_handlers.go` — every `GET /api/version` can trigger it (15-min cache); the web version badge polls that endpoint
- `cmd/octo-desktop/main.go` — `autoUpdateLoop` checks 30s after launch, then daily

No user data in the request, but the IP and the timing reach GitHub, and an air-gapped install just watched it fail on a timer.

## What

`update_check` in `~/.octo/config.yml`, default `true` — existing installs behave exactly as before — plus a Settings toggle wired through `PUT /api/config/update_check`.

Off silences both automatic paths. Explicit `octo upgrade`, `octo upgrade --check` and the tray's "Check for updates…" still reach out: running the command is the consent.

The preference is read fresh from config inside `latestVersion()` rather than captured into `server.Config`, so flipping the toggle takes effect on the next badge refresh instead of at the next restart. `server.Config.UpdateCheck` keeps its old job as the build-level gate that leaves the test suite and every non-`serve` constructor offline.

## Docs

README / README_CN / `landing/index.html` now say what octo actually does — one unprompted request, one switch away from silent — instead of claiming zero. The air-gapped guide points at the switch, and the key is in the config reference (en + zh).

## Tests

- `TestUpdateCheckEnabled_DefaultsOn` / `_RoundTrip` — default is on; `false` survives Save/Load (the `*bool` is load-bearing: `omitempty` would drop a plain `false` and silently re-enable it)
- `TestLatestVersion_ConfigOptOut` — with the preference off, the fake release origin records **zero** hits
- `TestPutUpdateCheck_PersistsAndSilences` — PUT persists, `GET /api/config` reflects it, and the very next `/api/version` sends nothing

`go test ./...` is clean apart from `TestCompressImageData_KeepsSmaller`, which fails identically on `main` with the local Go 1.27 toolchain. `go vet ./...`, `gofmt -l .`, the desktop module's vet, and `npm test` (686 tests) all pass.


Leihb:
## Review round — the new copy was falsified by the repo itself

Isolated review caught that the wording this PR introduced asserts something the codebase contradicts.

`web/index.html:11` (and the built `internal/server/webdist/index.html:11`, i.e. the copy embedded in the binary) loads the Iconify web component from `code.iconify.design`, and with no collection vendored locally the icon data is fetched from Iconify's public API too — 41 files use `<iconify-icon>`. So opening the Web UI or the desktop webview sends unprompted third-party requests, and `update_check` cannot silence them.

That is exactly the class of error this PR exists to fix, so the claims now say what octo actually does rather than promising zero. The air-gapped guide also gained the row it never had: icons render blank offline.

Other review fixes:

- **Preference read moved behind the TTL/backoff gate.** `/api/version` is unauthenticated and the cache is what stops a request flood becoming an outbound-request flood; a `config.Load()` ahead of it put a disk read and a YAML parse on every unauthenticated request. The switch still promises exactly what it did — no request is *sent* — a cached `latest` can just linger one TTL after flipping it.
- **Fail closed, not open.** Both paths used `err == nil && !enabled`, so an unparseable `config.yml` silently re-enabled the request the user had switched off. Server reads through `LoadCached` (keeps the last config that parsed); desktop skips the tick.
- PUT goes through `config.Mutate` rather than Load+Save, so a concurrent settings write can't be clobbered.
- Settings copy no longer claims the version check is the only outbound request, and drops backticks a plain-text span renders literally.
- `dev-docs/octo-upgrade-design.md` and `desktop-update-notice.md` describe the user-level switch.

New tests: explicit `octo upgrade --check` still reaches the network with the preference off — the "typing the command is the consent" half of the contract, previously guarded only by the absence of a check; the desktop gate extracted as `autoCheckAllowed` and covered including broken YAML; a broken config keeps the server silenced.

### Follow-up worth its own PR

Bundling the Iconify collection locally would fix both the privacy story and blank icons on isolated networks. Not folded in here — it's a build-pipeline change, not a docs/config one.

Leihb:
## Iconify is now bundled — the claim holds without a caveat

Folded into this PR rather than deferred, so the copy doesn't have to apologise for a dependency we were about to remove anyway.

`scripts/gen-icons.mjs` resolves every icon name the repo can produce into `src/lib/icons.generated.ts`, which `initIcons()` registers via `addCollection` before first paint. `index.html` loads nothing external. **114 icons, 60 KB of data, +82 KB on the bundle** before compression — against 1.3 MB if the full `ant-design` + `lucide` sets were carried wholesale.

It scans rather than reading a hand-kept list, because names are written in four places (markup, string constants the markup binds to, the curated expert profiles under `internal/agentprofile/defaults`, and one Go handler) and a list would rot the first time someone adds an icon. Candidates come from a deliberately loose regex and are validated against the real collections, so a false positive like `'text-align:center'` finds no icon and is dropped.

**User-authored icons.** An agent profile's `icon:` is YAML the user writes, so it can name something outside the bundle, and the component's reflex there is to ask the API. `initIcons` therefore also replaces iconify's `fetch`: no prefix, bundled or not, can reach the network. An unknown icon renders as nothing — which is what it did offline anyway.

So README / README_CN / landing go back to the strong claim, and the air-gapped guide loses the row about blank icons. Confirmed there is no `@font-face` and no other external `url()`; the iconify tag was the last external reference in the page shell.

### Found on the way

`FeedbackModal`'s unselected radio used `ant-design:circle-outlined`, which doesn't exist in that set — it has always rendered blank, CDN or not. Now `lucide:circle`.

### On the tests

The generated file is committed so dev and vitest work without running the generator, with a test asserting it still byte-matches a fresh run — a stale bundle fails silently, as a blank space, which is the exact failure mode this change exists to remove.

My first attempt at the no-network test spied on `globalThis.fetch` and **passed with the guard removed** — iconify captures its own fetch reference at module load, so replacing the global can't observe anything. Rewritten to compare that captured reference; verified red with the guard removed.

### New dependencies

`iconify-icon` (runtime — it is the same web component previously loaded from the CDN, now from `node_modules`); `@iconify-json/{ant-design,lucide,ph,logos,simple-icons}` and `@iconify/utils` as devDeps, used only by the generator. CI's `npm ci` installs devDeps and the `prebuild` hook regenerates; checked there is no `--omit=dev` anywhere.

Full suite: 49 Go packages pass (`TestCompressImageData_KeepsSmaller` fails identically on `main` with the local Go 1.27 toolchain); `npm test` 694 pass.

PR #2446, opened 2026-09-17
fix(desktop): one source of truth for "is there an update"

Leihb:
## Symptom

On desktop, the tray offers a download while the version badge in the bottom-left says you're up to date — same process, two answers.

## Why

They were never one thing. Four independent divergences, all verified in code:

| | 托盘 / tray | Web badge |
|---|---|---|
| state | `bridge.updateAvailable` (`cmd/octo-desktop/bridge.go:106`) | `s.versionLatest` + 15-min TTL |
| refresh | 30s after launch, then every 24h | once per page load |
| on failure | keeps its last answer → still offers the download | reports current-is-latest → goes quiet |
| budget | 10s | 3s |

The budget row is the one that bites deterministically. `upgrade.Check` walks GitHub plus four mirrors, giving each attempt a 5s sub-context **bounded by the caller's parent**. With a 3s parent, a slow (not refused) primary consumes the whole budget and all four mirrors receive an already-expired context — so the server could never reach `dl.octo-agent.dev` while the tray's 10s could. On a network where GitHub is slow, the tray finds the release and the badge never does. Every time.

## Fix

Collapse it to one fact:

- **The server's cache is the source of truth.** New `Server.LatestVersion()` / `Server.RefreshLatestVersion(ctx)`; the tray reads those instead of calling `upgrade.Check` itself. A manual tray check forces a fresh lookup *and seeds the cache*, so the badge agrees on its next read. Before the hub is bound (the 30s startup gap, or a failed takeover) the tray still falls back to a direct check — there's no badge to disagree with then.
- **Budget 3s → 10s**, so mirrors are actually reachable. One cold call can now take up to 10s; it's single-flighted and cached, so at most one client pays it per TTL and only the badge waits on it.
- **A failed lookup keeps the last release we saw** instead of downgrading to "current is latest". A lookup failing doesn't un-release a version — and this was the other half of the disagreement, since the tray already latched.
- **The badge re-checks every 15 minutes** (the cache TTL, so a tick inside the window costs no outbound request) rather than only on mount. A desktop window stays open for days; a mount-only check froze it.
- `upgrade.checkAttemptTimeout` becomes a `var` (like `BaseURL` already is) plus an exported `CheckAttemptWindow()`, so the budget relationship is testable in milliseconds instead of real seconds.

## Tests

- `TestCheck_SlowPrimaryStillReachesMirror` — a hanging primary, and the mirror still answers within the budget
- `TestCheck_BudgetShorterThanOneAttemptSkipsMirrors` — documents the failure mode so nobody lowers it back unaware: mirror hit count is **0**
- `TestVersionCheckBudgetCoversMoreThanOneAttempt` — cross-package invariant, `versionCheckTimeout > upgrade.CheckAttemptWindow()`
- `TestLatestVersion_FailureKeepsLastKnown` — a failed re-check, and one during backoff, both still report the known release
- `TestRefreshLatestVersion_BypassesCacheAndSeedsIt` — ignores the TTL, and the badge's next read matches
- `TestLatestVersion_ExportedMatchesBadge` — what the tray reads and what `/api/version` serves are equal
- `VersionBadge.test.ts` — picks up a release published after load; does not re-check mid-upgrade. Verified both fail with the interval removed.

## Note on #2445

That PR (the `update_check` privacy switch) touches the same two functions — `latestVersion()` and `runUpdateCheck()`. Whichever lands second needs a small rebase; the changes are complementary, not competing (the switch gates whether a lookup happens, this makes both surfaces read the same one).


Leihb:
## Review round — the fix didn't actually deliver sync

Isolated review turned up two things worth the rework.

**The shared cache only guarantees agreement at equal read times.** Each surface kept its answer in its own field until its next tick, so a tray on a daily cadence and a badge on a 15-minute one could still *display* contradictory things for most of a day. The systematic disagreement was fixed; the timing one wasn't.

**And the badge's interval was set equal to the cache TTL**, so every tick landed exactly past expiry and triggered a real lookup. The comment claiming a tick inside the window costs no outbound request was exactly backwards — the true cost was ~96 lookups/day.

So reading and refreshing are now separate concerns:

- `Server.LatestVersion` (what `/api/version` serves and what the tray calls) only reads the cache. A stale read starts one background refresh and answers with what's known. Reads are free.
- Refreshing has its own cadence — `versionRefreshInterval`, 6h.

Both surfaces re-read every minute, which is what actually keeps them showing one answer, while the network is touched **4 times a day instead of 96**. It also takes a lookup that can run the full 10s budget off the request path, where it was delaying the `native`/`local`/`os_version` fields the frontend reads from the same response. The lookup runs with no cache lock held, and its goroutine is joined in `doShutdown` instead of outliving the server (`-race` caught it reading `upgrade.BaseURL` during test teardown).

Also fixed: `needsUpdate` → `upgrade.NeedsUpdate` (the tray's no-hub fallback was a second copy of the rule, in the PR whose point was one rule); an auto tick no longer reaches the network when no hub is bound (at a minute's cadence that copy would have become a per-minute lookup); the tray menu only rebuilds when the answer changes; `RefreshLatestVersion` errors instead of claiming current-is-latest on a build that never looks; `checkVersion` re-checks phase after its await; every test that repoints `upgrade.BaseURL` now also clears `MirrorBaseURLs`.

`trayLookup` got the coverage it had none of. Full suite: 49 packages pass; `TestCompressImageData_KeepsSmaller` fails identically on `main` with the local Go 1.27 toolchain. `npm test` 689 pass.

### Pushed back on two review findings

- **"Reconnect budget shrinks 27s → 20s."** Not so — `VersionBadge.svelte:141-152` only checks the deadline at the top of each tick, and a successful fetch goes straight to `done` with no re-check. A slow `/api/version` makes the spinner linger, it doesn't cause a false `restart_failed`.
- **"10s worst case on the request path."** The severity was overstated: 10s is the budget ceiling, reached only if GitHub *and* `dl.octo-agent.dev` both hang. Still moved off the request path, but for the coupling, not the number.

Issue #2431, opened 2026-09-16
Per-model context_window for self-hosted endpoints (the table can match and still be wrong)


Leihb:
## Problem

`lookupContextWindow` (`internal/agent/compaction.go`) resolves a model's context window by substring-matching its name against a built-in table. The table records what a model's *published* ceiling is. For a self-hosted endpoint, the window is decided at deployment time and has nothing to do with the name:

- **It's sized against GPU memory at startup.** The same `Qwen3-32B` might be served with a 32k window on one box and 128k on another. The name is identical, so the table returns the same number for both.
- **The published ceiling often needs extra configuration to reach.** Qwen3-32B's `max_position_embeddings` is 40960; 131072 requires YaRN (factor 4.0) to be switched on explicitly. The table records the extended ceiling.
- **A gateway in front can truncate further** than the model itself.

The window drives the auto-compaction trigger, the Tool Search activation budget (`internal/tools/tool_search.go`) and the context gauge (`internal/server/handlers.go`, `ws_handlers.go`). When it over-estimates, turns get rejected by the endpoint and `internal/agent/overflow.go` recovers — once per turn, at the cost of a wasted round trip that reads to the user as an unexplained stall.

## Why `fallback_context_window` doesn't cover this

#2426 added `fallback_context_window` (flag / `OCTO_FALLBACK_CONTEXT_WINDOW` / config), but it is deliberately scoped to the fallback: it applies only when the name matches **no** table entry. That scoping is correct and shouldn't change — making it override known models would let one operator-supplied number flatten every correctly-known model in the process, including small ones like `llama-3` (8k) that it would inflate into overflow.

So a model named `my-internal-llm` is now fixable, and a model named `Qwen3-32B` is not — even when both are served from the same box with the same 32k window.

## Current workarounds, both bad

- **Rename the model** so it misses the table. Fragile: matching is by substring, so `qwen3-32b-internal` still hits `qwen3`.
- **Lower `compact_auto_pct`.** It's global, so every other model in the process starts compacting early too.

## Proposed

A per-model `context_window` on the endpoint's model entry, which the two-level config schema is already shaped for:

```yaml
endpoints:
  - id: intranet
    provider: custom
    protocol: openai
    base_url: http://10.0.0.20:8000/v1
    models:
      - model: Qwen3-32B
        context_window: 32000   # what --max-model-len was actually set to
```

Precedence: per-model entry > built-in table > `fallback_context_window` > built-in default.

[...]

PR #2447, opened 2026-09-17
fix(config): bind context windows to endpoint model deployments (#2431)

dolimcom:
Thank you for the review. The following changes address the issues you raised.

## Main changes

- Added `context_window` to each endpoint model entry.
- Updated context-window precedence to:
  `model-level context_window > built-in model table > fallback_context_window > 128k default`.
- Added a runtime-only `EndpointID` to `ModelEntry`; it is excluded from YAML and JSON serialization.
- Agent instances now retain the model, context window, and endpoint ID together, so deployment identity survives session recovery and model switching.
- Identically named models are bound to an exact endpoint through the `<endpoint>::<model>` composite ID.
- Automatic compaction, Tool Search, and the UI context gauge use the deployment window resolved for the Agent.
- Tool Search APIs now require `contextWindow`, preventing future callers from silently falling back to a bare-model lookup.
- The configuration API accepts `0` for automatic resolution and rejects values from `1` to `999`.
- Updated the English and Chinese documentation with precedence, token units, validation behavior, and air-gapped deployment examples.

### 1. Sub-agent model overrides lost the per-endpoint window

When a sub-agent selected another model on the parent's endpoint, the previous implementation only knew the parent's primary and lite windows. Any other sibling model fell back to the built-in table, so a deployment configured for `32000` tokens could incorrectly receive the built-in value, such as `131k`.

Agent instances now retain the current endpoint ID. When Spawner receives an explicit model override, it first resolves:

```text
<parent-endpoint>::<requested-model>
```

The child therefore receives the context window configured for the endpoint that will actually serve its requests, instead of resolving by bare model name. If the endpoint reference is stale or the configuration cannot be loaded, the existing bare-model fallback remains in place so sub-agent startup is not blocked.

A regression test covers two endpoints exposing the same sibling model: when the parent is bound to `endpoint-a`, the child receives that endpoint's `32000`-token window rather than `endpoint-b`'s `64000`-token window.

### 2. The web UI could erase a hand-configured `context_window`

The request DTO now represents `context_window` as `*int`, allowing the API to distinguish these cases:

- Field omitted: preserve the existing window, including values configured directly in `config.yml`.
- Explicit `0`: clear the deployment override and resume automatic window resolution.
- Value from `1` to `999`: return HTTP 400.

Tests cover preserving an omitted value, clearing with an explicit zero, and defaulting a newly added model to zero when the field is omitted. The behavior was also verified through real REST requests.

### 3. Other non-blocking suggestions

- The English and Chinese air-gapped guides now clarify that the API rejects values from `1` to `999`, while an invalid value written directly to `config.yml` is treated as unset at runtime and reported by `octo doctor`.
- The optional Tool Search window parameters are now required. Every current call site passes either the Agent's resolved window or an explicit `0`.
- Rate-limiting the repeated ambiguity warning for duplicated bare model names is not included in this change and is better handled as a separate follow-up.

## Verification

- Sub-agent sibling-model resolution on the parent's endpoint.
- API behavior for omitted, zero, invalid, and restored context-window values.
- Composite endpoint binding for same-name 32k and 64k deployments.
- Tool Search budgets and MCP bridge activation.
- Endpoint propagation through session recovery, Channel Agent, and Lite Model paths.
- A real DeepSeek session and chat request.
- `go vet`.
- Repository-wide Go compilation check.
- Frontend tests: `698 passed`.
- Frontend production build.
- Documentation site build.
- Targeted Go regression tests and repeated stability runs.

## Known limitation

Automatic compaction now uses the correct deployment window, but its trigger still uses:

```text
max(provider-reported tokens, heuristic token estimate)
```

Chinese text, system prompts, and tool schemas may be overestimated, so compaction can still occur before provider-reported usage reaches the configured percentage. Tokenization also differs across models and languages, which makes a single heuristic inherently approximate. This is an existing compaction-decision limitation rather than an endpoint-window binding issue.

If anything still needs adjustment, I would be happy to continue addressing it.


Leihb:
Thanks @dolimcom — this is exactly what we need for #2431. The window resolves from the endpoint entry at Agent construction instead of a bare-model table lookup, the precedence chain works as documented, and the CLI resume path correctly re-binds `entry` to the session's endpoint before `SetModelConfig`. The locking hygiene improvements are a nice bonus.

Two things block merge.

**1. Sub-agent model overrides lose the per-endpoint window.**

The agent tool advertises models available on the current endpoint, so `req.Model` can be a *sibling* model on the parent's endpoint. `ContextWindowFor` (`internal/agent/agent.go:699`) only knows the parent's primary and lite windows; anything else silently falls back to the built-in table. A sub-agent pinned to a model configured with `context_window: 32000` gets the table value instead — the exact bug class this PR fixes, alive on the sub-agent path. Pass a window resolver into `NewSpawner`, or document the limitation.

**2. The web UI can silently wipe a hand-configured `context_window`.**

`endpointModelIn.ContextWindow` is a plain `int` (`internal/server/onboard_config_handlers.go:1142`), so "omitted" and "explicit 0" are indistinguishable, and `UpsertModel` replaces the whole `EndpointModel`. The settings UI never sends the field (`web/src/components/settings/EndpointsSection.svelte:206,220`), so re-adding a model through the UI resets a config-file-set value to 0. Make the request field `*int` and preserve the existing value when nil.

**Smaller things**

- `internal/config/config.go:952`: `EntryByModel` runs per agent build/turn, so the ambiguity `slog.Warn` fires on every turn for a duplicated bare model name. Rate-limit or emit once per config load.
- The air-gapped docs (both locales) say values 1–999 are "rejected", but a hand-edited `config.yml` value under `octo serve` is silently treated as unset by `EffectiveContextWindow`. Worth one sentence.
- The variadic `contextWindow ...int` parameters (`internal/tools/registry.go`, `internal/tools/tool_search.go`) let a future caller omit the window and still compile, silently reintroducing the bare-name bug. A required parameter would make it compile-time enforced. (The only current omission is dead code at base — harmless today.)

Fix 1 and 2 and I'll merge it.

Leihb:
@dolimcom 加我微信，我拉你进群交流吧

dolimcom:
Thank you for the detailed review. I addressed both blocking items in `d2c84fda`:

- Sub-agent sibling-model overrides now resolve their context window against the parent's exact endpoint deployment.
- The endpoint-model API now distinguishes an omitted `context_window` from an explicit `0`, preserving hand-configured values while still allowing callers to clear an override.

I also made the Tool Search window argument required at all call sites and clarified the hand-edited invalid-value behavior in both air-gapped guides. Targeted regression tests, real REST/DeepSeek checks, `go vet`, frontend tests (`698 passed`), frontend build, docs build, and the repository-wide Go compilation check passed. The PR description has been updated with the implementation and remaining compaction limitation.

Leihb:
Both blockers are fixed and verified locally: the sibling-model test pins the child to endpoint-a's 32k window with endpoint-b exposing the same name at 64k, and the endpoint-model API now preserves an omitted `context_window` while an explicit `0` clears it. The required Tool Search window parameter and the hand-edited-value note in both guides are in as well.

I pushed one small commit on top (e0e52c7e) that removes `Agent.SetModelConfig` and `Agent.SetLiteModel`: after your second commit they had no production callers left, so there is now a single deployment-aware setter.

Merging. Thanks for the quick turnaround.

dolimcom:
> @dolimcom 加我微信，我拉你进群交流吧

好的，不过微信号是多少

PR #2461, opened 2026-09-17
fix(subagent): keep a cut-short sync sub-agent's work and make it queryable

Leihb:
A synchronous sub-agent stopped by the loop detector reached the parent as nothing but the stop notice, and `sub_agent_status` then rejected its id with *"only async sub-agents are tracked"*. The parent's only reasonable conclusion was that the child was gone. It wasn't — it was still in the spawner's live registry, resumable via `sub_agent_send` for another 30 minutes.

## Carry the work

`budgetStop` ends a `stuck` or `max_turns` round by **replacing** the model's text with its own notice, so everything the child produced lived only in its history. `runChild` now walks back for the last substantive assistant text and puts it in front of the notice:

```
[partial — the sub-agent's last message before it was cut off]
The flake is the 5s sub-context in upgrade.Check.

[octo] Stopped: detected repeated tool calls without progress. …

[INCOMPLETE: this sub-agent was stopped after repeating the same tool calls
without progress — the result above is only what it had produced by then. It
is still resumable with sub_agent_send (agent_id "dbb7aa4b") — continue it
with a DIFFERENT approach rather than the same instruction, or re-launch with
a narrower task.]
```

The text is labelled rather than presented as a result: the last thing a child said can be a finished summary or an opening "let me look" many rounds back, and the stop notice must not vouch for it. A child that only ever called tools has nothing to carry, and gets the bare notice.

`incompleteNote` is now shared by `sub_agent` and `sub_agent_send`, covers `stuck` as well as `max_turns`, and names the resume handle — the stop reason alone gave the model no reason to think the child was still there.

## Make it queryable

`Spawner` implements a new `tools.ChildInspector` (`InspectChild` / `ListChildren`) over its `childRegistry`, and `sub_agent_status` consults it when the manager doesn't know an id:

```
Sub-agent dbb7aa4b: idle, still resumable — send it a follow-up with
sub_agent_send using agent_id "dbb7aa4b".
It is not tracked as a background agent because it already returned its result
inline, but its context is intact for a follow-up.
Last round: stopped by the loop detector (repeated the same tool calls without
progress), 12 turn(s), 1m30s ago.
```

Details worth flagging for review:

- Lookups strip an `agent_` prefix on retry. Models read the `[agent dbb7aa4b]` tag as the `agent_N` form and ask for `agent_dbb7aa4b` — that exact mangling is what the reported failure hit. The reply always echoes the bare id `sub_agent_send` accepts.
- Registry reads go through a new `snapshot`, not `get`: a status query is a read and must not refresh LRU standing. Expired children are still reaped first, so an idle-expired id correctly reports as gone.
- The last-round fields on `liveChild` are guarded by their own `snapMu`. `lc.mu` is held for a whole round, so reusing it would make a status query block behind the running child.
- The id-less listing appends resumable children, de-duplicated against the manager's agents by the new `TrackedBackingIDs()` — an async spawn's child sits in both places and must be listed once. `ListRunning()` itself is untouched, since `pending_work` and the live panels read it.
- A spawner without the interface (test fakes) just has nothing to report; the tool falls back to its unknown-id error, whose wording no longer blames sync dispatch.

## Tests

`internal/app/spawner_stuck_test.go` drives a real child into the stuck detector and asserts the carry, the label ordering, the snapshot agreeing with what the parent got, and that a clean stop passes through untouched. `internal/tools/agent_status_sync_test.go` covers the sync lookup, the `agent_`-prefixed spelling, the listing (including no double-counting), the no-inspector spawner, and `incompleteNote`.

`make fmt-check`, `make vet`, and `go test -race` on `internal/app` + `internal/tools` all pass. The one failure on my machine, `TestCompressImageData_KeepsSmaller` in `internal/agent`, is a known local Go 1.27 image-encoder difference and is unrelated to this change.


Leihb:
An isolated review of the first commit found four defects in it, fixed in `96b330fc`. The branch is rebased onto current main (#2459 and #2460 both landed in `internal/app/spawner.go`).

**A running child was offered as resumable.** The registry holds a child from before its first round — `Spawn` registers it, then runs — while an async spawn's backing id isn't recorded on the manager until `Spawn` returns. Between those points `sub_agent_status` reported the same child twice: working under its `agent_N` handle, and again as a "finished, still resumable" bare id with zero turns. That bare id had never been exposed to the model before, and a follow-up sent to it would have blocked on the child's lock for the rest of the round. `runChild` now marks the child busy for the duration; a busy child is left out of the resumable listing, and a direct query reports it as working. De-duplicating by backing id alone can't cover this — the id is recorded later than the registration.

**The carry reached past its own round.** A round that produced no text walked back into the previous round and labelled *its* delivered answer as this round's partial work. The walk now stops at the round's prompt, distinguished from a tool-result message by its blocks rather than its role.

**A failed round left a stale snapshot.** `sub_agent_status` would present the previous round's result as the child's latest word. `setSnapErr` records the failure and clears the reply.

**`budgetStop` has three call sites, not two.** `max_tokens` ends a round with a synthetic notice exactly like `max_turns` and `stuck`, so it gets the same carry and the same `[INCOMPLETE]` note. The comment and the design doc claimed two.

Since the `[INCOMPLETE]` note is what makes a cut-short child legible, three related gaps went with them: the async completion note now shares `incompleteNote` instead of hard-coding `max_turns` (a stuck background child carried no marker at all), the TUI and web status words put `stuck` under incomplete/warning rather than reporting it completed, and `sub_agent_send` accepts the same `agent_`-prefixed spelling `sub_agent_status` does. Listings no longer copy every child's full reply only to discard it, and the retained reply is capped like the manager's.

Five new tests cover the busy child, the round boundary, the failed-round snapshot, the `max_turns` carry (previously only `stuck` was exercised end to end), and both listing sections together.

PR #2479, opened 2026-09-18
test(server): await version refresh in mustServer cleanup

### LINKED ISSUE #2457: flake: TestLatestVersion_ChecksAndCaches — upstream hits = 2, want 1
Seen once on the PR #2456 ubuntu-latest leg (run 35215125672, job 105181748159), `go test` with `-race`:

```
--- FAIL: TestLatestVersion_ChecksAndCaches (0.01s)
    version_upgrade_handlers_test.go:90: upstream hits = 2, want 1 (everything after the refresh is cached)
```

A rerun of the identical commit went green, and the PR's diff (cmd/octo TUI) cannot influence internal/server — so this is a pre-existing flake, not a regression from #2456.

## Analysis

`hits` counts requests to a per-test httptest fake, but `upgrade.BaseURL` is a **package-global** the test swaps (`version_upgrade_handlers_test.go:60-62`). A second hit therefore means some `upgrade.Check` ran while `BaseURL` pointed at this test's fake — and the failing test's own server can only fire one (the cold `LatestVersion` kicks a CAS-guarded background refresh; once it lands, the cache is fresh for 6h).

The likely source is a **stray background refresh from a previous test's server**: `mustServer`'s cleanup (`server_test.go:861`) drains turns and titles but never calls `srv.Shutdown` / `awaitVersionRefresh`, so a server whose test started a background refresh relies on the test itself calling `settleVersionRefresh`. Every `UpdateCheck: true` test in `version_upgrade_handlers_test.go` currently does settle — but settle only proves the goroutine's HTTP completed; any test (present or future) that starts a refresh after its settle, or forgets to settle, strands a goroutine whose `checkAndStore` reads the global `BaseURL` whenever it happens to run — under a slow `-race` CI runner that can be inside the *next* test's window, adding a hit to the wrong fake.

Two possible directions:

1. Test-side: make `mustServer`'s cleanup always `awaitVersionRefresh` (bounded), so no version goroutine can outlive its test regardless of what the test body did.
2. Structural: stop reading the package-global `upgrade.BaseURL` at call time inside `checkAndStore` — capture it per server (or per refresh start), so a stray goroutine hits its own test's fake (or the real GitHub) rather than whichever test swapped the global last.

Not urgent — one observed failure in one run — but the next occurrence will cost someone a confused CI rerun.

Repro attempt: `go test -race ./internal/server/ -run TestLatestVersion -count=50` (did not reproduce locally on a fast machine, consistent with a CI-load-dependent timing window).

Leihb:
## Problem

Fixes #2457 — `TestLatestVersion_ChecksAndCaches` flaked once on the PR #2456 ubuntu-latest leg with `upstream hits = 2, want 1`.

Root cause: `mustServer`'s cleanup drains turn and title goroutines but never waits on `versionRefreshWG`. A background version check (`startVersionRefresh`) that outlives its test calls `upgrade.Check`, which reads the package-global `upgrade.BaseURL` **at call time**. Under `-race` CI load that goroutine can fire after the next test has swapped the global, adding an upstream hit to the wrong per-test fake.

`settleVersionRefresh` can't prevent this: it only polls `versionChecking`, so a refresh started after a test's settle (or by a test that forgets to settle) escapes it.

## Fix

Test-side only, no production change: `mustServer`'s cleanup now awaits the in-flight version refresh, bounded at 10s.

Why this is race-free and sufficient:

- `versionRefreshWG.Add(1)` happens synchronously before the goroutine starts, so any refresh a test could have kicked is already counted when cleanup runs.
- `t.Cleanup` is LIFO: `mustServer`'s cleanup is registered after each test's `upgrade.BaseURL` restore, so it runs **before** the restore — a late goroutine still hits its own test's fake, never the next one.

## Verification

- `go build ./...`, `go vet ./internal/server/`
- `go test ./internal/server/ -run 'Version' -count=5`
- `go test -race ./internal/server/ -run 'TestLatestVersion' -count=20` (run twice, independently)


Leihb:
Superseded by #2462 (merged), which landed the same teardown await plus a delta-based assertion. Closing as redundant.

PR #2518, opened 2026-09-21
feat: let the agent see inside an open artifact, and hand a file back

Leihb:
An HTML artifact open in the panel is sealed off from the agent: it runs on its own origin behind a policy that closes every network exit. The agent wrote the page, but it has no idea what that page *renders as* — whether the layout holds, whether the chart drew, whether the contrast is legible. This gives the page a way to say so, and gives the agent a way to hand something back.

Three tools, scoped to the current session: `artifact_state` (every reporting page with its own one-line digest), `view_artifact` (its screenshot, into the conversation as a real image), `insert_into_artifact` (an image, into an open page).

## Why here, and not where it was

#2501 built exactly this machinery for Light Apps. The link worked; it was attached to the wrong thing. The flagship scenario — draw, generate, put it back — is a capability demo more than a workflow. The mirror needed the app's author to call `pushState`, so an app not written for it showed nothing when the user asked about it. And "the agent can only see it if you keep it open in a particular place" is a mental model that did not land (2026-09-21, the user's own words: 正常人无法理解). A Light App is a global tool that had been pinned into session context beside artifacts and the diff.

The need actually lives on session artifacts. The panel is their home already, the user discussing an artifact is by definition looking at it, and their author is the agent itself — teach it once in a skill and new artifacts carry the call, so there is no chicken-and-egg. The scenario the Light App version never had is the agent checking its own work: write the page, look at what it rendered, fix what is wrong.

So the Light App mirror retires here: its three tools, its `/api/light-apps/{slug}/state` and delivery endpoints, and the state and delivery halves of its bridge.

## The shape

The mirror is keyed by `(session_id, artifact_path)` — the identity an artifact already has. The tools read only the session stamped on the turn, so an agent never sees another conversation's panel.

A page publishes with one call, guarded so it also works opened bare:

```js
window.octo?.pushState({
  digest: 'A bar chart of Q3 sales, 8 bars.',  // one line, written for the model
  summary: { bars: 8 },                         // optional, any JSON
  image: canvasOrBlob,                          // optional screenshot
})
```

`lightapp_bridge.js` becomes `frame_bridge.js`: one script, `window.__octoBridge = {kind, ns, download?}`, with the kind selecting which halves boot — storage migration and the download proxy for a Light App, state and delivery for an artifact.

The page cannot reach the API from its origin, so the web UI relays: `PUT/DELETE /api/sessions/{id}/artifacts/state?path=…`, with the path checked against the session's transcript-derived whitelist, the same gate the preview endpoint enforces. Delivery mints a one-time ticket bound to `(session, path)` with a 2-minute TTL, redeemed over `GET /api/sessions/{id}/artifacts/delivery/{ticket}`; images only, and only the raster formats `NewImageBlock` can sniff, which keeps SVG — a document that carries script — out.

## What the page can and cannot do

A frame speaks only under the identity the host gave it at iframe creation, and only from the origin it was loaded from: a page can navigate itself away, and the WindowProxy survives that, so the registration is pinned to an origin rather than inherited by whatever lands next. Deliveries post to that origin, never `'*'`. The relayed summary is compacted, because JSON allows raw newlines between tokens and `artifact_state` renders one line per page. Snapshots are memory-only — replaced by the next push, stale at 5 minutes, evicted at 30, gone when the frame closes; only a screenshot actually handed to the model becomes part of the conversation.

An old Light App keeps `window.octo`, inert. The mirror was its feature first and the docs of the day taught the unguarded `window.octo.pushState(...)`, sitting in pointerup and change handlers where a TypeError would take the rest of the handler with it.

## User-visible impact

An **eye icon** in the panel's title bar when the artifact open there is publishing — the answer to "can the agent see this?", which is otherwise invisible from outside.

`lightapp_state`, `view_lightapp` and `insert_into_lightapp` are gone. A Light App's `publish()` is now a silent no-op rather than an error, and nothing else about Light Apps changes.

## Testing

- `internal/tools`: session isolation, the single-page default, the ambiguous case, no deliverer, missing arguments, eviction, staleness, summary truncation.
- `internal/server`: the relay's whitelist 404 and its PUT/DELETE round trip (asserted through the tool's real output, not the mirror's internals), summary line-forging, and delivery's five refusals — reused ticket, wrong session, expired, foreign artifact, non-image — plus `nosniff` and `CSP: sandbox` on the bytes.
- `web`: router identity checks, origin pinning both ways, double-host unregister, an in-flight PUT racing a DELETE, the delivery target, and the injected bridge's two shapes run as the bytes that ship.
- Full suites: Go packages pass, vitest 762/762 across 64 files, `svelte-check` 0 errors, docs site builds.
- Not done: no end-to-end pass in a real browser. The loop is covered by unit tests on both sides, not by having watched it happen.

Docs: a new `guides/artifacts` (EN + zh) in the sidebar — the panel had no page of its own, and this contract had nowhere to live once the Light App guide gave it up. `base.md`, the `artifact-design` skill and product-help's `WEB.md` follow. Design note: `dev-docs/artifact-interaction-design.md`.


Leihb:
Closing unmerged after hands-on testing: the artifact↔agent interaction is too awkward to use in practice. Publishing a snapshot and reading it back sounded like it removed a step, and in the hand it adds one — the page has to be written for it, opened in the right place, and kept open, and what comes back is a digest someone wrote in advance rather than an answer to what was asked.

That verdict applies to the idea, not to where it was attached, so the Light App mirror this PR was going to replace (#2501) has no case left either. It is being removed from main in a follow-up rather than kept as the surviving half. The design note goes with it.

The `mount: "panel"` retirement (#2519) stands on its own and is unaffected.

